PR #44, opened 2024-01-09
Mixnet: Add basic structure and topology construction

youngjoon-lee:
This is the first PR of the series of Mixnet spec.

This PR represents:
- https://www.notion.so/Mixnet-Specification-807b624444a54a4b88afa1cc80e100c2#1e3a0deec22242d3a12cb54b65413049
- https://www.notion.so/Mixnet-Specification-807b624444a54a4b88afa1cc80e100c2?pvs=4#18b6e714554447d5a824aec1f1891403

### DIFF
--- a/mixnet/fisheryates.py
+++ b/mixnet/fisheryates.py
@@ -4,22 +4,18 @@ from typing import List
 
 class FisherYates:
     @staticmethod
-    def sample(elements: List, count: int, entropy: bytes) -> List:
+    def shuffle(elements: List, entropy: bytes) -> List:
         """
-        A sampling algorithm using Fisher-Yates shuffling.
-        In Python, random.sample uses Fisher-Yates shuffling internally by default.
-        https://softwareengineering.stackexchange.com/a/215780
-        https://docs.python.org/3/library/random.html#random.shuffle
-
-        If FisherYates shuffling is not provided by a certain programming language by default,
-        a developer has to implement FisherYates shuffling that can be used for sampling.
+        Fisher-Yates shuffling algorithm.
+        In Python, random.shuffle implements the Fisher-Yates shuffling.
         https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle
-        :param elements: elements that will be used for sampling
-        :param count: the number of elements to be sampled
+        https://softwareengineering.stackexchange.com/a/215780
+        :param elements: elements to be shuffled
         :param entropy: a seed for deterministic sampling
         """
+        out = elements.copy()
         random.seed(a=entropy, version=2)
-        sampled = random.sample(elements, count)
+        random.shuffle(out)
         # reset seed
         random.seed()
-        return sampled
+        return out
--- a/mixnet/fisheryates.py
+++ b/mixnet/fisheryates.py
@@ -1,15 +1,25 @@
+import random
 from typing import List
-
-from numpy.random import Generator
-from randomgen.chacha import ChaCha
 
 
 class FisherYates:
     @staticmethod
-    def shuffle(elements: List, entropy: bytes) -> List:
-        out = elements.copy()
-        rng = Generator(ChaCha(seed=list(entropy)))
-        for i in reversed(range(1, len(out))):
-            j = rng.integers(low=0, high=i, endpoint=True)  # [low,high]
-            out[i], out[j] = out[j], out[i]
-        return out
+    def sample(elements: List, count: int, entropy: bytes) -> List:
+        """
+        A sampling algorithm using Fisher-Yates shuffling.
+        In Python, random.sample uses Fisher-Yates shuffling internally by default.
+        https://softwareengineering.stackexchange.com/a/215780
+        https://docs.python.org/3/library/random.html#random.shuffle
+
+        If FisherYates shuffling is not provided by a certain programming language by default,
+        a developer has to implement FisherYates shuffling that can be used for sampling.
+        https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle
+        :param elements: elements that will be used for sampling
+        :param count: the number of elements to be sampled
+        :param entropy: a seed for deterministic sampling
+        """
+        random.seed(a=entropy, version=2)
+        sampled = random.sample(elements, count)
+        # reset seed
+        random.seed()
+        return sampled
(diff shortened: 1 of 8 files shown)


mixnet/fisheryates.py
```diff
@@ -0,0 +1,15 @@
...
+
+class FisherYates:
+    @staticmethod
+    def shuffle(elements: List, entropy: bytes) -> List:
+        out = elements.copy()
+        rng = Generator(ChaCha(seed=list(entropy)))
+        for i in reversed(range(1, len(out))):
+            j = rng.integers(low=0, high=i, endpoint=True)  # [low,high]
+            out[i], out[j] = out[j], out[i]
+        return out
```
danielSanchezQ, line 15: Fisher yates is the default python shuffler btw. Probably not worth to maintain it in specs and would be simplers. We have a wrapper in the carnot package you can take a look at (We can copy paste that instead, better not to have inter package dependencies).

mixnet/mixnet.py
```diff
@@ -0,0 +1,66 @@
...
+    ) -> MixnetTopology:
+        assert n_nodes_per_layer * n_layers < len(self.mix_nodes)
+
+        shuffled = FisherYates.shuffle(self.mix_nodes, entropy)
+        sampled = shuffled[: n_nodes_per_layer * n_layers]
+        layers = []
+        for l in range(n_layers):
+            start = l * n_nodes_per_layer
+            layer = sampled[start : start + n_nodes_per_layer]
+            layers.append(layer)
```
danielSanchezQ, line 40: I have the feeling this could be a `pick_layer` or `sample_layer` method.

mixnet/mixnet.py
```diff
@@ -0,0 +1,66 @@
...
+        cls,
+        identity_private_key: BlsPrivateKey,
+        encryption_private_key: X25519PrivateKey,
+        addr: NodeAddress,
+    ) -> Self:
+        return cls(
+            identity_private_key.get_g1(),
+            encryption_private_key.public_key(),
+            addr,
+        )
```
danielSanchezQ, line 61: Why is `build` a *classmethod* instead of a simple constructor (`__init__`) here?

mixnet/mixnet.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+@dataclass
+class MixnetTopology:
+    layers: List[List[MixNode]]
```
danielSanchezQ, line 66: Nitpick: empty line at the end of file

mixnet/fisheryates.py
danielSanchezQ: Maybe something like?:

```python
class FisherYates:
    @staticmethod
    def shuffle(elements: List, entropy: bytes) -> List:
        """
       Fisher-yates shuffling algorithm
       It is the one used by python by default
       https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle
       https://softwareengineering.stackexchange.com/a/215780
       :param elements:
       :param entropy:
       :return:
       """
        out = elements.copy()
        random.seed(a=entropy, version=2)
        random.shuffle(out)
        # reset seed
        random.seed()
        return out
```

mixnet/mixnet.py
youngjoon-lee: True. If we don't need to maintain the default constructor (auto-generated `__init__` by `@dataclass) along with the `build`, we can simply implement `__init__` instead of `build`. In next PRs, there are some cases which needs two constructors. But here, we don't need two. I'll change it.

mixnet/mixnet.py
youngjoon-lee: I think it's because of the difference between python formatters. I use [black](https://pypi.org/project/black/) for formatting and [isort](https://pypi.org/project/isort/) for sorting imports. Which formatter do you use? We can make a sync.

mixnet/fisheryates.py
youngjoon-lee: Oh, I didn't know that the `random.shuffle` is actually the Fisher-Yates.

Also, I found that we can use `random.sample` instead of using `random.shuffle` and implementing sampling by ourselves, as described in https://docs.python.org/3/library/random.html#random.shuffle. I don't think this abstraction doesn't reduce readability of the specification. 
705dfeb994ef2e08e1e6b4927dc3b3f73fcb6ca8

mixnet/mixnet.py
youngjoon-lee: I just tried to extract this as a function, but the function was not neat as I expected because the function needs an assumption that the length of array must be always `n_nodes_per_layer * n_layer`.
If the implementation above is not very complicated, I think it's better to have it because it shows the full context about how many nodes were sampled.

mixnet/mixnet.py
youngjoon-lee: Done 0a649c7

mixnet/mixnet.py
danielSanchezQ: I think we could probably standarice [ruff](https://github.com/astral-sh/ruff) as formatter and linter. It's built in **rust**!!

mixnet/fisheryates.py
danielSanchezQ: Yes, `random.sample` can be used. But we need to know what sampling algorithm it uses as it needs to be clear in the specification.

danielSanchezQ: Oh, just saw in the commit that sampel it actually uses *fisher_yates* internally. Neat!

```
A sampling algorithm using Fisher-Yates shuffling.
        In Python, random.sample uses Fisher-Yates shuffling internally by default.
        https://softwareengineering.stackexchange.com/a/215780
        https://docs.python.org/3/library/random.html#random.shuffle
```

danielSanchezQ (approved):

mixnet/fisheryates.py
youngjoon-lee: Ah, I just looked at the internal source code of `random.sample`. It seems that it doesn't use FisherYates sampling internally. I had guessed it from the [sentence](https://docs.python.org/3/library/random.html#random.shuffle): 
>  random.shuffle(x)
>
>    Shuffle the sequence x in place.
>
>    To shuffle an immutable sequence and return a new shuffled list, use sample(x, k=len(x)) instead.

, but I think it's better to use `random.shuffle` explicitly and implement sampling as I implemented at the beginning.

PR #49, opened 2024-01-15
Mixnet: Packet delay in mix node

youngjoon-lee:
This PR covers the specification of packet delay in mix nodes: https://www.notion.so/Mixnet-Specification-807b624444a54a4b88afa1cc80e100c2?pvs=4#15ccbb03a44f42658ea67a33527ecea2.

In summary,
assuming that the packet receiving rate of a mix node follows a Poisson distribution $\lambda$, the mix node should be implementation to approximate a [M/M/∞ queue](https://en.wikipedia.org/wiki/M/M/%E2%88%9E_queue) with packet delays that follow an exponential distribution $\mu$. With this setup, the packet emission rate of a mix node must be $\lambda$, and the mean number of packets being processed by a mix node at any time must be $\lambda/\mu$. 
A unit test checks these properties.

### DIFF
--- a/mixnet/node.py
+++ b/mixnet/node.py
@@ -27,8 +27,10 @@ NodeId: TypeAlias = BlsPublicKey
 # 32-byte that represents an IP address and a port of a mix node.
 NodeAddress: TypeAlias = bytes
 
-InboundSocket: TypeAlias = "queue.Queue[SphinxPacket]"
-OutboundSocket: TypeAlias = "queue.Queue[Tuple[NodeAddress, SphinxPacket | Payload]]"
+PacketQueue: TypeAlias = "queue.Queue[Tuple[NodeAddress, SphinxPacket]]"
+PacketPayloadQueue: TypeAlias = (
+    "queue.Queue[Tuple[NodeAddress, SphinxPacket | Payload]]"
+)
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


mixnet/node.py
```diff
@@ -0,0 +1,174 @@
...
+    ) -> MixNodeRunner:
+        thread = MixNodeRunner(
+            self.encryption_private_key,
+            delay_rate_per_min,
+            inbound_socket,
+            outbound_socket,
+        )
+        thread.daemon = True
+        thread.start()
+        return thread
```
danielSanchezQ, line 65: Nitpick: could this be async instead?

danielSanchezQ (commented):
I think in general is ok. I would change the use of threads by async code instead. Also in previous PRs I remember to see some async code, we should choose one style. But it is up to you.

mixnet/node.py
danielSanchezQ: > I haven't used async code in this repo. But, hm, let me translate this into async. If it doesn't look complicated, we can go with async because it's much easier to approximate M/M/∞ queue using async rather than sync.

It is not necessary!, we can go ahead with this.

youngjoon-lee: Yeah, but after opening subsequent PRs #50 and #51, now I have a feeling that async code would be even simpler. I'll try it and open a separate PR soon. So, I hope this and subsequent PRs are reviewed and merged independently.

PR #50, opened 2024-01-16
Mixnet: mix client emission

youngjoon-lee:
This PR covers [Poisson packet emission from Nomos nodes (mix clients)](https://www.notion.so/Mixnet-Specification-807b624444a54a4b88afa1cc80e100c2?pvs=4#b0c91c5a8fb84414870df8998891f3b4).

### DIFF
--- a/mixnet/client.py
+++ b/mixnet/client.py
@@ -1,0 +1,75 @@
+from __future__ import annotations
+
+import queue
+import time
+from datetime import datetime, timedelta
+from threading import Thread
+
+from mixnet.mixnet import Mixnet, MixnetTopology
+from mixnet.node import PacketQueue
+from mixnet.packet import PacketBuilder
+from mixnet.poisson import poisson_interval_sec
+
+
+class MixClientRunner(Thread):
+    """
+    Emit packets at the Poisson emission_rate_per_min.
+
+    If a real packet is scheduled to be sent, this thread sends the real packet to the mixnet,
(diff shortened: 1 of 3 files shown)


mixnet/client.py
```diff
@@ -0,0 +1,75 @@
...
+        self.topology = topology
+        self.emission_rate_per_min = emission_rate_per_min
+        self.redundancy = redundancy
+        self.real_packet_queue = real_packet_queue
+        self.redundant_real_packet_queue: PacketQueue = queue.Queue()
+        self.outbound_socket = outbound_socket
+
+    def run(self) -> None:
+        # Here in Python, this thread is implemented in synchronous manner.
+        # In the real implementation, consider implementing this in asynchronous if possible.
```
danielSanchezQ, line 44: Yes, here I see clearly that this is meant to be a long running task. Is this where you [were referring ](https://github.com/logos-co/nomos-specs/pull/49#discussion_r1457117373) about changing to async?

danielSanchezQ (approved):
Looking pretty good and self contained. Thanks!

mixnet/client.py
youngjoon-lee: Yes, exactly. This thread has many short sleeps to minimize the time distortion(?), compared to having a long sleep, such as `time.sleep(poisson_interval_sec(self.emission_rate_per_min)`, because there might be some computation time to process packets. I actually don't like this approach. Also, I guess this can be more readable if we use async timer.

PR #51, opened 2024-01-18
Mixnet: Periodical topology update + Test refactoring

youngjoon-lee:
This PR is the last piece of Mixnet specification. This covers the [deterministic periodical topology reconstruction](https://www.notion.so/Mixnet-Specification-807b624444a54a4b88afa1cc80e100c2?pvs=4#af141ade3e8147ec900cf3599953f4da).

For the simplicity, this Python spec doesn't describe how network connections are established and maintained. Instead, this focuses on showing how entropy is provided and used to build topologies. Managing network connections is an implementation detail.

### DIFF
--- a/mixnet/fisheryates.py
+++ b/mixnet/fisheryates.py
@@ -4,11 +4,7 @@ from typing import List
 
 class FisherYates:
     @staticmethod
-    def set_seed(seed: bytes) -> None:
-        random.seed(a=seed, version=2)
-
-    @staticmethod
-    def shuffle(elements: List) -> List:
+    def shuffle(elements: List, seed: bytes) -> List:
         """
         Fisher-Yates shuffling algorithm.
         In Python, random.shuffle implements the Fisher-Yates shuffling.
@@ -17,6 +13,9 @@ class FisherYates:
         :param elements: elements to be shuffled
         :param entropy: a seed for deterministic sampling
         """
+        random.seed(a=seed, version=2)
         out = elements.copy()
         random.shuffle(out)
+        # reset seed
+        random.seed()
         return out
--- a/mixnet/fisheryates.py
+++ b/mixnet/fisheryates.py
@@ -4,7 +4,11 @@ from typing import List
 
 class FisherYates:
     @staticmethod
-    def shuffle(elements: List, entropy: bytes) -> List:
+    def set_seed(seed: bytes) -> None:
+        random.seed(a=seed, version=2)
+
+    @staticmethod
+    def shuffle(elements: List) -> List:
         """
         Fisher-Yates shuffling algorithm.
         In Python, random.shuffle implements the Fisher-Yates shuffling.
@@ -14,8 +18,5 @@ class FisherYates:
         :param entropy: a seed for deterministic sampling
         """
         out = elements.copy()
-        random.seed(a=entropy, version=2)
         random.shuffle(out)
-        # reset seed
-        random.seed()
         return out
--- a/mixnet/mixnet.py
+++ b/mixnet/mixnet.py
@@ -2,20 +2,10 @@ from __future__ import annotations
 
 import random
 from dataclasses import dataclass
-from typing import List, TypeAlias
+from typing import List
 
-from cryptography.hazmat.primitives.asymmetric.x25519 import (
-    X25519PrivateKey,
-    X25519PublicKey,
-)
-from pysphinx.node import Node
-
-from mixnet.bls import BlsPrivateKey, BlsPublicKey
 from mixnet.fisheryates import FisherYates
-
-NodeId: TypeAlias = BlsPublicKey
-# 32-byte that represents an IP address and a port of a mix node.
-NodeAddress: TypeAlias = bytes
+from mixnet.node import MixNode
 
 
 @dataclass
@@ -47,22 +37,6 @@ class Mixnet:
 
     def choose_mixnode(self) -> MixNode:
         return random.choice(self.mix_nodes)
-
-
-@dataclass
-class MixNode:
-    identity_private_key: BlsPrivateKey
-    encryption_private_key: X25519PrivateKey
-    addr: NodeAddress
-
-    def identity_public_key(self) -> BlsPublicKey:
-        return self.identity_private_key.get_g1()
-
-    def encryption_public_key(self) -> X25519PublicKey:
-        return self.encryption_private_key.public_key()
-
-    def sphinx_node(self) -> Node:
-        return Node(self.encryption_private_key, self.addr)
 
 
 @dataclass
--- a/mixnet/mixnet.py
+++ b/mixnet/mixnet.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
+import queue
 import random
 import threading
 import time
-import queue
 from dataclasses import dataclass
 from datetime import datetime, timedelta
 from threading import Thread
@@ -11,7 +11,6 @@ from typing import List, TypeAlias
 
 from mixnet.fisheryates import FisherYates
 from mixnet.node import MixNode
-
 
 EntropyQueue: TypeAlias = "queue.Queue[bytes]"
 
@@ -26,7 +25,7 @@ class Mixnet:
         n_layers: int,
         n_nodes_per_layer: int,
         current_entropy: bytes,
-        entropy_delay_sec: int,
+        topology_update_delay_sec: int,
     ):
         self.mix_nodes = mix_nodes
 
@@ -36,7 +35,7 @@ class Mixnet:
             n_layers,
             n_nodes_per_layer,
             current_entropy,
-            entropy_delay_sec,
+            topology_update_delay_sec,
             self.entropy_queue,
         )
         self.topology_updater.daemon = True
@@ -59,14 +58,14 @@ class MixnetTopologyUpdater(Thread):
         n_layers: int,
         n_nodes_per_layer: int,
         current_entropy: bytes,
-        entropy_delay_sec: int,
+        topology_update_delay_sec: int,
         entropy_queue: EntropyQueue,
     ):
         super().__init__()
         self.mix_nodes = mix_nodes
         self.n_layers = n_layers
         self.n_nodes_per_layer = n_nodes_per_layer
-        self.entropy_delay_sec = entropy_delay_sec
+        self.topology_update_delay_sec = topology_update_delay_sec
         self.entropy_queue = entropy_queue
 
         self.lock = threading.Lock()
@@ -76,12 +75,15 @@ class MixnetTopologyUpdater(Thread):
     def run(self) -> None:
         while True:
             new_entropy = self.entropy_queue.get(block=True)
-            ts = datetime.now() + timedelta(seconds=self.entropy_delay_sec)
+            ts = datetime.now() + timedelta(seconds=self.topology_update_delay_sec)
 
             next_topology = self.build_topology(new_entropy)
             self.establish_conections(next_topology)
 
-            time.sleep((ts - datetime.now()).total_seconds())
+            remain = (ts - datetime.now()).total_seconds()
+            if remain > 0:
+                time.sleep(remain)
+
             with self.lock:
                 self.topology = next_topology
 
--- a/mixnet/mixnet.py
+++ b/mixnet/mixnet.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
 
 import random
+import threading
+import time
 from dataclasses import dataclass
+from datetime import datetime, timedelta
+from threading import Thread
 from typing import List
 
 from mixnet.fisheryates import FisherYates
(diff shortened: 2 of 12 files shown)


mixnet/fisheryates.py
```diff
@@ -4,7 +4,11 @@
 
 class FisherYates:
     @staticmethod
-    def shuffle(elements: List, entropy: bytes) -> List:
+    def set_seed(seed: bytes) -> None:
+        random.seed(a=seed, version=2)
+
+    @staticmethod
+    def shuffle(elements: List) -> List:
```
danielSanchezQ, line 11: I think the old style is better. IIRC from the expects, the entropy changes over time. In this case we would set the whole system seed when calling `set_seed`, which is not what we want. It should be reset on every call to maintain integrity with other random calls. (This is python specific, but it actually fixates the intention that the entropy is just for the shuffling).

mixnet/mixnet.py
```diff
@@ -12,31 +16,105 @@
...
+
+    def set_entropy(self, entropy: bytes) -> None:
+        """
+        Set entropy (seed) to the Fisher-Yates shuffler
+        that is going to be used for deterministic topology construction.
+
+        This method is expected to be called periodically by the user.
+        Until the next entropy is provides, Fisher-Yates shuffler continues to works
+        based on this entropy set once.
+        """
```
danielSanchezQ, line 46: Unnecessary wrapper IMO

mixnet/mixnet.py
```diff
@@ -12,31 +16,105 @@
...
+
+        self.lock = threading.Lock()
+        self.topology = self.build_topology()
+        self.next_topology = None
+
+    def run(self) -> None:
+        next_epoch_ts = datetime.now() + timedelta(seconds=self.mixnet_epoch_sec)
+
+        while True:
+            time.sleep(1 / 1000)
```
danielSanchezQ, line 80: Why is there a sleep here? nitpick: I think schedules are defined already when checking it below. Probably can be abstracted into some scheduler. 
Not necessary to change now. But we could have a blackboxed structure than awaits till next epoch.

mixnet/mixnet.py
```diff
@@ -12,31 +16,105 @@
...
+        self.n_nodes_per_layer = n_nodes_per_layer
+        self.mixnet_epoch_sec = mixnet_epoch_sec
+        self.proactive_sec = proactive_sec
+
+        self.lock = threading.Lock()
+        self.topology = self.build_topology()
+        self.next_topology = None
+
+    def run(self) -> None:
+        next_epoch_ts = datetime.now() + timedelta(seconds=self.mixnet_epoch_sec)
```
madxor, line 77: Thinking at loud here.

I don't know if this is the best way to count epochs, it may work here when all node instances have their clocks synchronized. However, when one node will connect later (or will disconnect for some time) he might loose track of time here. It might be better to have something like `self.genesis_timestamp` from which all nodes calculated current epoch:
`epoch_number = ceiling(datetime.now() - params.genesis_timestamp / params.epoch_len_in_sec)`
and slot number as (relative to the epoch):
`slot_number_in_epoch = ceiling((datetime.now() - epoch_number * params.epoch_len_in_sec) / params.slot_len_in_sec)`
Or something similar.

This way the calculation should be independent from the connectivity conditions of nodes and can be influenced only by clock drifts.

Another things is that the epoch and slot information should be confirmed by reading them from the consensus. But that might be out of the scope for this specification effort.

mixnet/mixnet.py
```diff
@@ -12,31 +16,105 @@
...
+
+        while True:
+            time.sleep(1 / 1000)
+
+            now = datetime.now()
+            if now < next_epoch_ts - timedelta(seconds=self.proactive_sec):
+                continue
+
+            if self.next_topology is None:
+                self.next_topology = self.build_topology()
```
madxor, line 87: If I understand it correctly, in this version, we are refreshing the topology only on once per epoch?

mixnet/mixnet.py
```diff
@@ -12,31 +16,105 @@
...
+        self.next_topology = None
+
+    def run(self) -> None:
+        next_epoch_ts = datetime.now() + timedelta(seconds=self.mixnet_epoch_sec)
+
+        while True:
+            time.sleep(1 / 1000)
+
+            now = datetime.now()
+            if now < next_epoch_ts - timedelta(seconds=self.proactive_sec):
```
madxor, line 83: I hope that I understand this correctly, the intention here is to update the topology before the next epoch, so that when the new epoch starts it can already use the new topology?

If so, then it might be problematic due to the construction of the random beacon. That is, the beacon stability is more probable with time, that is, it is better to wait longer for the beacon. In other words, can we consider that the new topology must be usable with some delay, for example: the beacon is "stable" when it is in calculated in slot 1 of every epoch, then the topology is updated with some delay for example we assume that it is calculated in slot 2 and, usable for slot 3 and later of each epoch. Therefore, we are giving more time for network synchronization.

How does it sounds?

Also for me it is easier to reason when I we switch from thinking in seconds into thinking in slots and epochs. I see less synchronization problems ;)

mixnet/mixnet.py
```diff
@@ -12,31 +16,105 @@
...
+    def current_topology(self):
+        with self.lock:
+            return self.topology
+
+    def build_topology(
+        self,
     ) -> MixnetTopology:
-        num_nodes = n_nodes_per_layer * n_layers
+        num_nodes = self.n_nodes_per_layer * self.n_layers
         assert num_nodes < len(self.mix_nodes)
```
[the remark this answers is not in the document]
danielSanchezQ: good catch!

mixnet/fisheryates.py
youngjoon-lee: Previously, yes, the entropy changes over time. But, at the recent [discussion](https://www.notion.so/Mixnet-Specification-807b624444a54a4b88afa1cc80e100c2?d=e35faabaf784477f9a7ea3e393ccb071&pvs=4#6a9d677c11b34ab7a3a00299c3c196e6), we've changed the design. The entropy is going to be injected less frequently, for example, 5 days. But, the topology update is going to be triggered more frequently, e.g. 30 minutes. So, the seed, which was set when the entropy was injected, should be maintained over subsequent topology updates until a new entropy is injected.

mixnet/mixnet.py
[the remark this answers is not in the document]
youngjoon-lee: Thank you! Yes, it can be equal. I'll fix this.

mixnet/mixnet.py
```diff
@@ -12,31 +16,105 @@
...
 
-        shuffled = FisherYates.shuffle(self.mix_nodes, entropy)
+        shuffled = FisherYates.shuffle(self.mix_nodes)
         sampled = shuffled[:num_nodes]
         layers = []
-        for l in range(n_layers):
-            start = l * n_nodes_per_layer
-            layer = sampled[start : start + n_nodes_per_layer]
+        for l in range(self.n_layers):
```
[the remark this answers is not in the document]
youngjoon-lee: Yes. I agree. Technically we don't actually need to split the `sampled` array into several subarrays. Here, I did it deliberately for readability.

mixnet/mixnet.py
youngjoon-lee: Yes. I missed this point. I'll fix it.

> Another things is that the epoch and slot information should be confirmed by reading them from the consensus. But that might be out of the scope for this specification effort.

I'm also thinking about this. Here, we only use `mixnet_epoch_sec` that defines the interval of topology updates, which is going to be shorter than `consensus_epoch_sec`. So, if we design the entropy to be injected from the consensus layer to the mixnet layer, I think the mixnet layer doesn't need to read something (consensus epoch or slot info) from the consensus layer. Did I understand the point that you mentioned?

mixnet/mixnet.py
youngjoon-lee: Yes. The topology is updated once per `mixnet_epoch` (e.g. 30min), which is going to be shorter than `consensus_epoch` (e.g. 5d). Also, the `mixnet_epoch` is going to be independent with `consensus_slot` (20sec). That might be longer.

mixnet/mixnet.py
youngjoon-lee: The strategy you suggested is reasonable if we design topology to be updated once per `consensus_epoch`. But, this PR assumed that the topology is updated once per `mixnet_epoch`, which is shorter than `consensus_epoch`, because I thought the `consensus_epoch` is going to be very long (e.g. 5days). I thought this is what we've discussed in Notion, but I probably didn't understand it correctly. Let me think which way between these two approach is more reasonable.

youngjoon-lee: Ah, even if we have `mixnet_epoch` shorter than `consensus_epoch`, we also need to use the newly injected entropy with some delays, as you explained.

mixnet/fisheryates.py
youngjoon-lee: I think this is the topic that we need to make a sync in https://github.com/logos-co/nomos-specs/pull/51#discussion_r1457675581

danielSanchezQ: Yes. I think in this case there is a missunderstanding here. `set_seed` should be instead `set_entropy` then we store the entropy and do not set the seed inmidiately. Then we use the old code. Otherwise is gonna be impossible to make simulations for example.

youngjoon-lee: Yes. If you remove the `mixnet_epoch` concept. We can use the old code. Also, not setting the seed immediately looks reasonable (https://github.com/logos-co/nomos-specs/pull/51#discussion_r1457694044).

mixnet/mixnet.py
madxor: I think that it might be easier to think about mixnet functions as a tool that will be used by the consensus. In my head, the consensus should be responsible for triggering mixnet function, as the consensus should be treated as the source of truth. That is, what if we focus only on providing an API that potentially could be used as part of the consensus who will be responsible for synchronization of time and data, and the mixnet API will only respond (by setting the connectivity) accordingly to data provided by consensus. This way we could simplify a lot the internal logic of the mixnet and relieve it from knowing anything about the time or consensus.

The API could be very simple, such like:
`topology = get_mixnet_topology(`nodes_list`, entropy)` <-- returns topology configuration based on the entropy and the list of nodes.
`set_mixnet_topology(topology)` <-- activates the topology.

I think that mixnet should only care about the two above, and all synchronization/configuration issues should be abstracted from the mixnet and done on the consensus side. Therefore, the consensus be responsible for calculating and providing synchronized data in a timely manner to the mixnet.

Technically, I think that there should be another layer on top of the consensus that would do the above in conjunction to parsing governance transactions (coordination layer?), through which some of the global parameters will be set. But we don't have it yet so I'm using consensus here.

mixnet/mixnet.py
madxor: Yes, I think that we are thinking the same way here. I've left more in depth comment below in the next conversation (https://github.com/logos-co/nomos-specs/pull/51#discussion_r1459185382)

mixnet/mixnet.py
youngjoon-lee: > That is, what if we focus only on providing an API that potentially could be used as part of the consensus who will be responsible for synchronization of time and data,

Exactly. This is the same as what I thought at https://github.com/logos-co/nomos-specs/pull/51#discussion_r1458613901. But, the two APIs that you suggested are the more abstracted version than my recent fix https://github.com/logos-co/nomos-specs/commit/9a64242f4596f9b5d78aa103ad002d8394b1ad5f (linked in in https://github.com/logos-co/nomos-specs/pull/51#discussion_r1458613901). In general, I prefer your API because the mixnet doesn't need to care about the `entropy_delay_sec` (in my recent fix).

The last thing we need to take into account when designing these two APIs is that we also want for the mix node to establish connections with adjacent mix layer in the new topology [in advance](https://github.com/logos-co/nomos-specs/commit/9a64242f4596f9b5d78aa103ad002d8394b1ad5f#diff-4b6fc539badd8395558d0df89c79dfa602639423cb4b9f91593249c733a4ba6eR82). Maybe we can design mix node to start connection establishment in background as soon as the new topology is returned from `get_mixnet_topology` API. Unless the consensus layer calls `set_mixnet_topology` in very short time, the mix node can have time to establish connections in advance.

danielSanchezQ: I'm not so sure this should be like that. Consensus should know nothing of the comunication layer behind. It just knows there is something where it can send messages and receive messages. Eventually the mixnet can be even opt-in for that and should not matter. This being said, I think consensus can trigger some events of what is happening in the system (we get to decide that) then mixnet can use of that. But as possible, consensus and mixnet is better they do not know each other or the do know as little as possible.

youngjoon-lee: From https://github.com/logos-co/nomos-specs/pull/51#discussion_r1458613901, I removed the concept `mixnet_epoch`. But previously, yes, mixnet epoch was much shorter than consensus epoch. 

Now, I agree that having mixnet epoch is little complicated. Instead we can remove mixnet epoch, and make mixnet update topology whenever a new entropy is injected from the upper layer. Even, mixnet doesn't need to care about the frequency of entropy injection. Also, consensus layer doesn't need to know deeply about what would happen in the mixnet. The one thing that consensus layer has to do is propagating the entropy updated by it to other layers, which is something like a logical clock, as you said.

I think this decouples mixnet and consensus quite well, and this is what https://github.com/logos-co/nomos-specs/commit/9a64242f4596f9b5d78aa103ad002d8394b1ad5f and https://github.com/logos-co/nomos-specs/pull/51#discussion_r1459185382 try to achieve.
I think that you, Marcin, and I am already on the same page with this point. Did I understand your comment correctly?

youngjoon-lee: @madxor @danielSanchezQ I think we're almost on the same page and we can now come to a conclusion.

I made a commit 9a64242f4596f9b5d78aa103ad002d8394b1ad5f for clarity and another commit 1a970641a8847f8695e4c3f0cafb48ab32c3971f to rename a variable. These commit has API different from what Marcin suggested in https://github.com/logos-co/nomos-specs/pull/51#discussion_r1459185382, but I'm going to explain about this at the end of this comment.

In these commit, the consensus layer just injects a new entropy when it thinks that the new entropy is almost safe to use. Potentially, the consensus layer will be able to have some delays before injecting the entropy, instead of injecting it as soon as the random beacon is updated, because of https://github.com/logos-co/nomos-specs/pull/51#discussion_r1457694044. But anyway, it's the responsibility of the consensus layer. Mixnet layer doesn't need to know about this detail.

In perspective of mixnet, it just uses a new entropy whenever it is injected. Mixnet doesn't need to care about the frequency of entropy injection. It is up to the consensus layer. Even if the new entropy is injected every 20 seconds, it's okay. Mixnet will just use it to build a new topology, as soon as it arrives in. 

The one thing Mixnet cares about is putting some delay before switching to the new topology. As mentioned in https://github.com/logos-co/nomos-specs/pull/51#discussion_r1459244787, it takes some time for each mix node to establish new connections according to the new topology. Also, there may be some timing differences on the new entropy being injected. Of course, we can ignore this, by allowing a chaotic period in which packets are propagated through unexpected paths. But, I think we can minimize this chaotic period by putting some delay (e.g. 1min) before switching to the new topology. Then, each mix node can earn 1min to calculate the new topology and establish new connections. This is what 1a970641a8847f8695e4c3f0cafb48ab32c3971f tries to describe. This delay is not perfect, but at least, it can minimize the chaotic transition in the network that cannot be fully synchronized.

As I mentioned above, I designed API different from what Marcin suggested, because I didn't want to expose the concept of "topology" to the consensus layer by providing `get/set_topology` API. In that case, the consensus layer may be confused about how long it should hold the `topology` returned from `get_topology(entropy)` before calling `set_topology(topology)`. Instead of that, I thought it's simpler to have a single API `inject_entropy(entropy)`, so that the consensus layer just injects a new entropy whenever it wants and forgets about that.

What do you think about this? We have had many valuable discussions, and I think we are in the end of the discussion.

youngjoon-lee: I'm just attaching a diagram that helps our discussion. The option 2 is what my code is suggesting, but let's discuss today.
mixnet-topology-update drawio
Even if we choose the option 2, the first topology generation (when the process is started) should be like the option 1 probably.

madxor: I think that we are on the right path. However, I think we are still missing another layer to make it perfect. I fully agree that there should be a logical separation between the mixnet (and its time-awareness) and consensus (and its network awareness). Therefore, it might be necessary to create another layer on top of the mixnet and consensus, we can call it robustness (or privacy) agent/layer (or something). The robustness layer would be responsible for reading consensus and setting mixnet, we could configure it separately from the mixnet and consensus. Therefore, the flow would always be from the consensus event to mixnet action with independently configured logic.

For example, if change the length of the consensus epoch or slot time, then we should not adjust mixnet delay, that should be taken care by the robustness abstraction, which would translate the consensus timings to the mixnet ones (in practice, I imagine that mixnet will not be aware on any specific timings but will react on robustness calls).

danielSanchezQ: I agree as well. We need something in the middle to coordinate efforts. I wan to understand more about the delays as I am getting a bit lost on that. I feel the responsibility is kind of mixed. This system should be totally reactive IMO.

youngjoon-lee: This is a summary of our discussion today. 

mixnet-topology-update drawio(3)

- We're gonna have a separate "robustness" layer on top of consensus layer and mixnet layer.
- Robustness layer is gonna know parameters for building mixnet topologies.
- Whenever a new entropy is stabilized in the consensus layer, the consensus layer is going to emit the new entropy to the robustness layer, along with the updated list of nodes in the network. In general, this is going to happen once per consensus epoch, but it depends on the design of consensus layer, strictly speaking.
- As soon as a new information arrives in, the robustness layer "immediately" builds a new topology configuration based on the current parameter that it knows. The topology configuration means the list of mix nodes for each layer (for now in v1).
- As soon as the new topology configuration is generated, the robustness layer injects it to the mixnet layer. There is going to be no time delay between those two events.
- The mixnet layer switches the topology configuration immediately, and starts establishing new QUIC connections in background.
  - Packets based on the new topology can arrive in the mix node, even before all new connections are established. But, it is totally okay. Mix node always establish a new connection on demand to forward the packet with the best effort. Even if packets based on unexpected topology is arrived in, mix node never reject packets. This design is reasonable because we assume that all nodes in the network are ready to act as a mix node (at least for v1).

youngjoon-lee:
Closing this, and I'm going to open another PR according to https://github.com/logos-co/nomos-specs/pull/51#discussion_r1463179784.

PR #52, opened 2024-01-19
Add header id and message format specification

zeegomo:
An initial sketch for the format of consensus messages.
For the moment, the contents are completely opaque, and there's no distinction between cl transactions and bl blobs, the consensus ignores those in the current spec anyway

To clarify, this is just the definition of how messages should look like on the network, so that different implementation have a common ground.


danielSanchezQ:
Not that I complain. But if we later are gonna need it in python, wouldn't it be better to just have it directly? Or is this complementary so we can link it to documents?

zeegomo:
> Not that I complain. But if we later are gonna need it in python, wouldn't it be better to just have it directly? Or is this complementary so we can link it to documents?

This is a definition of the formats of messages, it does not have to be executable, there's no 'action' it's describing. To me, using an ad-hoc DSL feels more natural than writing code for serialization / deserialization.

danielSanchezQ:
> > I'm down for it. But just to remember we would have to probably maintain both.
> 
> This is the only specification of what the network message format should be and the sole source of truth. If we will write some code to serialize from/to this format it's just going to be an implementation of what is written in this doc

Makes sense, I misunderstood the purpose. Thanks for the explanation!

cryptarchia/messages.abnf
```diff
@@ -0,0 +1,19 @@
+; VERSION 0.1
+
```
davidrusu, line 2: Do we need one more rule for tying it all together?
i.e.
```abnf
BLOCK = HEADER CONTENT
```

cryptarchia/messages.abnf
```diff
@@ -0,0 +1,21 @@
+; VERSION 0.1
+; ------------ BLOCK ----------------------
+BLOCK            = HEADER CONTENT
+; ------------ HEADER ---------------------
+VERSION          = %x01
+HEADER           = VERSION HEADER-UNSIGNED
+HEADER-UNSIGNED  = %x00 HEADER-COMMON
+HEADER-COMMON    = CONTENT-SIZE CONTENT-ID BLOCK-DATE PARENT-ID
+CONTENT-SIZE     = U32
+BLOCK-DATE       = BLOCK-SLOT
```
zeegomo, line 10: should we use epoch + slot or just absolute slot? any consideration?

davidrusu: epoch seems useful to know.

One potential use: say you receive a block, you want to know if it's relevant to you or not right? just knowing the slot and the parent wouldn't be enough to know if your just behind? or if this block is actually from a completely different epoch that we've forgotten about.

cryptarchia/cryptarchia.py
```diff
@@ -46,8 +46,30 @@ class Config:
...
+        bytes += int.to_bytes(self.content_size, length=4, byteorder="big")
+        # content id
+        assert len(self.content_id) == 32
+        bytes += self.content_id
+        # slot
+        bytes += int.to_bytes(self.slot.absolute_slot, length=8, byteorder="big")
+        # parent
+        assert len(self.content_id) == 32
+        bytes += self.parent
+        return blake2b(bytes, digest_size=32).digest()
```
davidrusu, line 72: I believe the correct way to do this sort of hashing is to feed bytes to the hasher object instead of concatenating all bytes together and then hashing. e.g.

```python3
h = blake2b(digest_size=32)
h.update(b"\x01") # version byte
h.update(b"\x00") # header type for block
# ...
return h.digest()
```

cryptarchia/messages.abnf
```diff
@@ -0,0 +1,21 @@
...
+PARENT-ID        = HEADER-ID
+
+; ------------ CONTENT --------------------
+CONTENT          = *OCTET
+
+; ------------- MISC ----------------------
+U32              = 4OCTET  ; unsigned integer 32 bit (BE)
+U64              = 8OCTET  ; unsigned integer 32 bit (BE)
+HEADER-ID        = 32OCTET
+CONTENT-ID       = 32OCTET
```
davidrusu, line 21: missing new-line

cryptarchia/test_fork_choice.py
```diff
@@ -8,26 +8,31 @@
 
 
 def make_block(parent_id: Id, slot: Slot, block_id: Id) -> BlockHeader:
```
davidrusu, line 10: rename block_id to content_id now that we are using it for content?

cryptarchia/test_fork_choice.py
```diff
@@ -8,26 +8,31 @@
...
+        ]
         long_chain = deepcopy(common)
         short_chain = deepcopy(common)
 
         for slot in range(50, 100):
             # make arbitrary ids for the different chain so that the blocks appear to be different
-            long_id = hashlib.sha256(f"{slot}-long".encode()).digest()
-            short_id = hashlib.sha256(f"{slot}-short".encode()).digest()
+            long_id = hashlib.sha256(f"{slot}-long".encode()).digest()[:32]
+            short_id = hashlib.sha256(f"{slot}-short".encode()).digest()[:32]
```
davidrusu, line 28: Why do we need the `[:32]` on the end here?

cryptarchia/test_fork_choice.py
davidrusu: also, I think it would clean up the tests a bit if you have `make_block()` take `content` as input instead of `content_id`. Then here you can hash the content to get the content id for the header.

cryptarchia/cryptarchia.py
```diff
@@ -46,8 +46,30 @@ class Config:
...
+        bytes += b"\x00"
+        # content size
+        bytes += int.to_bytes(self.content_size, length=4, byteorder="big")
+        # content id
+        assert len(self.content_id) == 32
+        bytes += self.content_id
+        # slot
+        bytes += int.to_bytes(self.slot.absolute_slot, length=8, byteorder="big")
+        # parent
+        assert len(self.content_id) == 32
```
youngjoon-lee, line 70: I think we don’t need this assertion here again because we have it in line 65

cryptarchia/messages.abnf
davidrusu: actually, I take that back I'm looking at the ouroborous byron specification and I don't see an epoch listed. Probably wise to keep things as minimal as we can.
Screenshot 2024-01-31 at 2 27 25 PM

zeegomo: > just knowing the slot and the parent wouldn't be enough to know if your just behind? or if this block is actually from a completely different epoch that we've forgotten about.

epoch + slot inside the epoch or absolute slot are equivalent in term of information, I feel the latter is a easier to work with for time calculations but a bit more inconvenient for snapshots and such.
Anyway, this is just one format, what is used in the code for other operations is entirely decoupled.

cryptarchia/cryptarchia.py
zeegomo: curious about the difference

cryptarchia/test_fork_choice.py
zeegomo: the digest is by default 64bytes, I could not find an easy way to set that except  constructing a sha256 instance (which this convenience method does under the hood)

cryptarchia/cryptarchia.py
zeegomo: yes, good catch

cryptarchia/test_fork_choice.py
davidrusu: I'm not seeing 64 bytes here?
```python
>>> len(sha256(b"blah").digest())
32
```
In this looks like a no-op
```python
>>> hash = sha256(b"blah").digest()
>>> hash[:32] == hash
True
```

cryptarchia/cryptarchia.py
davidrusu: they produce the same result, but using the update api is what you will see in practice everywhere. Main reason is avoiding unnecessary allocations while building up the bytes buffer

cryptarchia/test_fork_choice.py
zeegomo: looks like you're right, what was I looking at ?🤔

davidrusu (approved):
🚢

PR #53, opened 2024-01-23
Mixnet: Refactor with asyncio

youngjoon-lee:
As previously discussed in https://github.com/logos-co/nomos-specs/pull/49#discussion_r1453154141, https://github.com/logos-co/nomos-specs/pull/50#discussion_r1457568051, https://github.com/logos-co/nomos-specs/pull/51#discussion_r1457710293, and https://github.com/logos-co/nomos-specs/pull/51#discussion_r1457573764, I refactored mixnet specs with using [asyncio](https://docs.python.org/3/library/asyncio.html). Now, it's more readable and simpler than the previous code that had verbose time calculations and sleeps.

### DIFF
--- a/mixnet/node.py
+++ b/mixnet/node.py
@@ -1,10 +1,7 @@
 from __future__ import annotations
 
-import queue
-import threading
-import time
+import asyncio
 from dataclasses import dataclass
-from threading import Thread
 from typing import Tuple, TypeAlias
 
 from cryptography.hazmat.primitives.asymmetric.x25519 import (
@@ -27,9 +24,9 @@ NodeId: TypeAlias = BlsPublicKey
 # 32-byte that represents an IP address and a port of a mix node.
 NodeAddress: TypeAlias = bytes
 
-PacketQueue: TypeAlias = "queue.Queue[Tuple[NodeAddress, SphinxPacket]]"
+PacketQueue: TypeAlias = "asyncio.Queue[Tuple[NodeAddress, SphinxPacket]]"
 PacketPayloadQueue: TypeAlias = (
-    "queue.Queue[Tuple[NodeAddress, SphinxPacket | Payload]]"
+    "asyncio.Queue[Tuple[NodeAddress, SphinxPacket | Payload]]"
 )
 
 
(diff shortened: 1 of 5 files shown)


mixnet/node.py
```diff
@@ -48,28 +45,28 @@ def encryption_public_key(self) -> X25519PublicKey:
...
             self.encryption_private_key,
             delay_rate_per_min,
             inbound_socket,
             outbound_socket,
         )
-        thread.daemon = True
-        thread.start()
-        return thread
+        task = asyncio.create_task(runner.run())
+        return runner, task
```
danielSanchezQ, line 61: Uhm, if we create a task this do not really need to be `async` as there is no `await`.

mixnet/node.py
```diff
@@ -79,96 +76,59 @@ def __init__(
         inbound_socket: PacketQueue,
         outbound_socket: PacketPayloadQueue,
     ):
-        super().__init__()
         self.encryption_private_key = encryption_private_key
         self.delay_rate_per_min = delay_rate_per_min
         self.inbound_socket = inbound_socket
         self.outbound_socket = outbound_socket
-        self.num_processing = AtomicInt(0)
+        self.packet_processing_tasks = set()
```
danielSanchezQ, line 83: What is this used for?

youngjoon-lee: This is just for keeping track of the number of tasks being processed, so that, in the test, we can check how many packets are staying in a mix node: https://github.com/logos-co/nomos-specs/pull/53/files#diff-b773b3f0f891e8bb9864e3aa1dbbef792c2f6cfcd0a7d9e4fce70f7ea6745598R127. In real implementations, we actually don't need this. I'll leave a comment in the code.

mixnet/node.py
youngjoon-lee: Absolutely. I fixed it: bd02fb4

mixnet/node.py
danielSanchezQ: I think this then should be part of just the testing code. It is a bit confusing if we mix specification with testing.

youngjoon-lee: Done c7b32fe. Now we don't have the `set` as a class field. Also, all calculations for testing are done within tests.

danielSanchezQ (approved):

PR #55, opened 2024-01-24
feat(lottery): spec out basic leader slot check

davidrusu:
In this PR:
- basic mock VRF for use in the slot leader checks
- introduces LedgerState, this holds a snapshot of the ledger state at some block height
- introduces EpochState, this holds the various epoch values used in the current epoch
- statistical test on the leader slot lottery to ensure it does actually give reasonable results.

I've also added [black](https://pypi.org/project/black/) to the requirements and have run it on the `cryptarchia/` directory. We can use this to auto-format our code.


cryptarchia/cryptarchia.py
```diff
@@ -1,21 +1,34 @@
...
+    chain_start_time: int
+
+
 # An absolute unique indentifier of a slot, counting incrementally from 0
 @dataclass
 class Slot:
     absolute_slot: int
 
-    def from_unix_timestamp_s(config: TimeConfig, timestamp_s: int) -> Date:
+    def from_unix_timestamp_s(config: TimeConfig, timestamp_s: int) -> "Slot":
```
zeegomo, line 31: what are the quotes for?

cryptarchia/cryptarchia.py
```diff
@@ -78,30 +83,29 @@ def __init__(self, genesis: BlockHeader, config: Config):
 
     # We don't do any validation in the current version
     def validate_header(block: BlockHeader) -> bool:
-        True
+        return True
```
zeegomo, line 86: whoopsie, rust to python transition problems

cryptarchia/cryptarchia.py
```diff
@@ -118,37 +122,103 @@ def on_block(self, block: BlockHeader):
...
 
-    def is_leader_at(self, slot: Slot) -> bool:
-        for coin in self.coins:
-            if lottery(slot, coin):
-                return True
-        return False
-    
     def propose_block(self, slot: Slot, parent: BlockHeader) -> BlockHeader:
         assert self.is_leader_at(slot)
         return BlockHeader(parent=parent.id(), slot=slot, _id=Id(b"TODO"))
```
zeegomo, line 220: `_id` has been removed?

cryptarchia/cryptarchia.py
```diff
@@ -118,37 +122,103 @@ def on_block(self, block: BlockHeader):
...
     value: int
 
+
+@dataclass
+class LedgerState:
+    """
+    A snapshot of the ledger state up to some height
+    """
+
+    height: int = None
```
zeegomo, line 146: height is not a unique identifier of ledger state

cryptarchia/cryptarchia.py
```diff
@@ -118,37 +122,103 @@ def on_block(self, block: BlockHeader):
...
+
+    # The nonce snapshot is taken 7k/f slots into the previous epoch
+    nonce_snapshot: LedgerState
+
+    def stake_distribution(self) -> int:
+        """Returns the total stake that will be used to reletivize leadership proofs during this epoch"""
+        return self.stake_distribution_snapshot.total_stake
+
+    def nonce(self) -> bytes:
+        return self.nonce_snapshot.nonce
```
zeegomo, line 167: are we going to need something more complicated than the total stake or were total stake requires some calculations? If not, can we simplify this and remove `LedgerState`? We probably can't take a stake distribution snapshot anyway if coins are private

cryptarchia/cryptarchia.py
```diff
@@ -118,37 +122,103 @@ def on_block(self, block: BlockHeader):
...
+    # for details of snapshot schedule please see:
+    # https://github.com/IntersectMBO/ouroboros-consensus/blob/fe245ac1d8dbfb563ede2fdb6585055e12ce9738/docs/website/contents/for-developers/Glossary.md#epoch-structure
+
+    # The stake distribution snapshot is taken at the beginning of the previous epoch
+    stake_distribution_snapshot: LedgerState
+
+    # The nonce snapshot is taken 7k/f slots into the previous epoch
+    nonce_snapshot: LedgerState
+
+    def stake_distribution(self) -> int:
```
zeegomo, line 162: nit: `stake_distribution` is not a good name for something returning total stake

cryptarchia/cryptarchia.py
davidrusu: You can't use refer to the same class name in a type signature, this is a limitation of python's interpreter in that it evaluates the class body before it the class symbol is ready for use.

cryptarchia/cryptarchia.py
davidrusu: Right, let me remove that line, something about the underscore in `_id` was not playing well with python dataclasses.

Besides, I think block ID should be a derived field of the block header.

cryptarchia/cryptarchia.py
davidrusu: true, perhaps a block id instead.

cryptarchia/cryptarchia.py
davidrusu: Whatever data is derived from the snapshots is not too important right now, the important thing is that they are old enough to be used in the current epoch safely and that the snapshots are taken in the correct order.

I.e. stake distribution snapshot is taken before nonce snapshot to prevent grinding.

> are we going to need something more complicated than the total stake or were total stake requires some calculations?

There will be more to this as we expand the leader proofs spec. For example we need to prove that a coin was present when the stake snapshot was taken.

>  If not, can we simplify this and remove LedgerState?

As I was reading the ouroborous glossary, it seemed like the idea of ledger state is quite fundamental, for example the genesis is not a block but a genesis ledger state: https://github.com/IntersectMBO/ouroboros-consensus/blob/fe245ac1d8dbfb563ede2fdb6585055e12ce9738/docs/website/contents/for-developers/Glossary.md#ledger-state

cryptarchia/cryptarchia.py
davidrusu: Agreed, I borrowed the name from ouroborous, do you think it's worth changing to `total_stake` or shall we keep in line with their terminology?

cryptarchia/cryptarchia.py
zeegomo: > Besides, I think block ID should be a derived field of the block header.

yes, it's just that we don't have enough data for now

cryptarchia/cryptarchia.py
zeegomo: Do they actually call stake distribution the total stake?

cryptarchia/cryptarchia.py
zeegomo: We're definitely going to need this at some point, just wondering whether it's best to introduce it when we actually need it.

> the important thing is that they are old enough to be used in the current epoch safely and that the snapshots are taken in the correct order.

true

zeegomo: > the idea of ledger state is quite fundamental

I think you convinced me it may be useful to have this concept from the start even though at the moment it's not particularly populated

cryptarchia/cryptarchia.py
davidrusu: https://github.com/IntersectMBO/ouroboros-consensus/blob/fe245ac1d8dbfb563ede2fdb6585055e12ce9738/docs/website/contents/for-developers/Glossary.md#stake-and-delegation

Reading this again, no I don't think that stake distribution == total stake, I'll rename this function 👍

PR #56, opened 2024-01-25
Mixnet: topology update

youngjoon-lee:
This is a new PR replacing the previous PR #51 closed.

This PR describes a [deterministic topology update](https://www.notion.so/Mixnet-Specification-807b624444a54a4b88afa1cc80e100c2?pvs=4#af141ade3e8147ec900cf3599953f4da) by introducing a "robustness layer" between a consensus layer and a mixnet layer. For details of design rationales, please see the spec in Notion that I linked above.

### DIFF
--- a/mixnet/mixnet.py
+++ b/mixnet/mixnet.py
@@ -10,14 +10,12 @@ from mixnet.node import MixNode
 class Mixnet:
     __topology: MixnetTopology | None = None
 
-    @property
-    def topology(self) -> MixnetTopology:
+    def get_topology(self) -> MixnetTopology:
         if self.__topology is None:
             raise RuntimeError("topology is not set yet")
         return self.__topology
 
-    @topology.setter
-    def topology(self, topology: MixnetTopology) -> None:
+    def set_topology(self, topology: MixnetTopology) -> None:
         """
         Replace the old topology with the new topology received, and start establishing new network connections in background.
 
--- a/mixnet/mixnet.py
+++ b/mixnet/mixnet.py
@@ -19,15 +19,14 @@ class Mixnet:
         Here in the spec, this method has been simplified as a setter, assuming the single-thread test environment.
         """
         self.topology = topology
-        self.establish_connections(topology)
+        self.establish_connections()
 
     def get_topology(self) -> MixnetTopology:
         if self.topology is None:
             raise RuntimeError("topology is not set yet")
         return self.topology
 
-    @staticmethod
-    def establish_connections(_: MixnetTopology) -> None:
+    def establish_connections(self) -> None:
         """
         Establish network connections in advance based on the topology received.
 
--- a/mixnet/mixnet.py
+++ b/mixnet/mixnet.py
@@ -4,44 +4,59 @@ import random
 from dataclasses import dataclass
 from typing import List
 
-from mixnet.fisheryates import FisherYates
 from mixnet.node import MixNode
 
 
 @dataclass
 class Mixnet:
-    mix_nodes: List[MixNode]
+    topology: MixnetTopology | None = None
 
-    # Build a new topology deterministically using an entropy.
-    # The entropy is expected to be injected from outside.
-    #
-    # TODO: Implement constructing a new topology in advance to minimize the topology transition time.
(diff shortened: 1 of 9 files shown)


mixnet/mixnet.py
```diff
@@ -4,44 +4,59 @@
...
+        self.topology = topology
+        self.establish_connections(topology)
+
+    def get_topology(self) -> MixnetTopology:
+        if self.topology is None:
+            raise RuntimeError("topology is not set yet")
+        return self.topology
+
+    @staticmethod
+    def establish_connections(_: MixnetTopology) -> None:
```
danielSanchezQ, line 30: nitpic: shouldn't this be a method with self instead?

youngjoon-lee: Absolutely. Thank you. This is a leftover from when I tried to make it async with some dummy implementation. But now, as you can said, I think it's enough to have a comment here instead of something complicated. I just modified it to a method with `self`: d145107

mixnet/robustness.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Given a entropy received, build a new topology and send it to mixnet.
+
+        In real implementations, this method should be a long-running task, consuming entropy periodically.
+        Here in the spec, this method has been simplified as a setter, assuming the single-thread test environment.
+        """
+        topology = self.build_topology(entropy)
+        self.mixnet.topology = topology
+
+    def build_topology(self, entropy: bytes) -> MixnetTopology:
```
madxor, line 49: Shouldn't somewhere here a "set_topology" that configures to the mixnet?

mixnet/mixnet.py
```diff
@@ -4,44 +4,59 @@
...
+    # In production, this can be a 1-D array, which is accessible by indexes.
+    # Here, we use a 2-D array for readability.
     layers: List[List[MixNode]]
 
     def generate_route(self) -> list[MixNode]:
         return [random.choice(layer) for layer in self.layers]
+
+    def choose_mix_destionation(self) -> MixNode:
+        all_mixnodes = [mixnode for layer in self.layers for mixnode in layer]
+        return random.choice(all_mixnodes)
```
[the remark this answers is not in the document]
youngjoon-lee: In this code, yes, the destination is selected by the mixnet layer automatically within the set of all mix nodes in the topology. It’s because the only information that the mixnet layer is the topology.
But I just remembered that we’ve discussed to choose a destination from the full set of nodes (including nodes not selected as mix). 
The robustness layer knows the full set. But, when the consensus layer or other layers try to send a message using mixnet, I think they will directly interact with the mixnet layer without asking the robustness layer.

Actually, after writing this Python code, now I think we can actually use mixnodes in the exit layer as mix destination, as you mentioned at the end of your comment. That will save one hop. Design is gonna be simpler. And, I don’t see any downside yet in terms of security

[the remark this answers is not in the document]
youngjoon-lee: If we use the exit layer as mix destinations, maybe adversaries may know that all packets coming into a node F in the exit layer is associated with some of them. That could be a downside if there is not much traffic in the network. But I think this information can be also leaked if someone can track egresses of the exit layer. Also in Nym, they have the concept of ‘service provider’ where all associated packets are gathered to. Therefore, I agree with this strategy. If you think so, I’ll update this and Notion

mixnet/robustness.py
youngjoon-lee: Updated e121220

mixnet/mixnet.py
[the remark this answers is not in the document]
youngjoon-lee: > Therefore, I agree with this strategy. If you think so, I’ll update this and Notion

@madxor I've thought about it. I think it's reasonable to use the last mix layer for mix destinations, instead of introducing an extra hop. I just updated [Notion](https://www.notion.so/Mixnet-Specification-807b624444a54a4b88afa1cc80e100c2?pvs=4#ebb923fa6a7443a0a36737906104343c), and Python code: 2aa9d5d. I added a new [diagram](https://www.notion.so/Mixnet-Specification-807b624444a54a4b88afa1cc80e100c2?pvs=4#ebb923fa6a7443a0a36737906104343c) in Notion, so anyone can understand it clearly.

danielSanchezQ (approved):

PR #57, opened 2024-01-26
Mixnet: integrate all the pieces together

youngjoon-lee:
Now that we have a clear view of the entire mixnet structure, this PR integrates all the components together, in order to clarify what are public APIs and what are private methods. 

There have been some refactoring. So, I recommend to read a README first that I added in this PR.

To make code-review easily, I left some comments below.


mixnet/mixnet.py
```diff
@@ -1,62 +1,58 @@
...
-        This is just a preparation to forward subsequent packets as quickly as possible,
-        but this is not a strict requirement.
-
-        In real implementations, this should be a background task.
-        """
-        pass
-
-
-@dataclass
-class MixnetTopology:
```
youngjoon-lee, line 43: Moved to `topology.py`

mixnet/node.py
```diff
@@ -17,71 +15,50 @@
...
 PacketPayloadQueue: TypeAlias = (
     "asyncio.Queue[Tuple[NodeAddress, SphinxPacket | Payload]]"
 )
 
 
-@dataclass
 class MixNode:
-    identity_private_key: BlsPrivateKey
-    encryption_private_key: X25519PrivateKey
-    addr: NodeAddress
```
youngjoon-lee, line 37: Moved to `topology.py` with a new name `MixNodeInfo`.

The previous name `MixNode` is now used by `node.py`, which defines the behavior of mix node.

mixnet/test_mixnet.py
```diff
@@ -1,40 +1,27 @@
...
-from mixnet.mixnet import Mixnet, MixnetTopologySize, MixNode
+from mixnet.mixnet import Mixnet
 from mixnet.robustness import Robustness
-from mixnet.utils import random_bytes
+from mixnet.test_utils import init
 
 
 class TestMixnet(IsolatedAsyncioTestCase):
-    @staticmethod
-    def init() -> Tuple[Mixnet, Robustness]:
```
youngjoon-lee, line 14: Moved to `test_utils.py`

mixnet/client.py
```diff
@@ -1,76 +1,122 @@
...
+    __real_packet_queue: PacketQueue
+    __outbound_socket: PacketQueue
+    __task: asyncio.Task  # A reference just to prevent task from being garbage collected
+
+    @classmethod
+    async def new(
+        cls,
+        initial_topology: MixnetTopology,
+        emission_rate_per_min: int,  # Poisson rate parameter: lambda in the spec
+        redundancy: int,  # b in the spec
```
madxor, line 25: Just an idea, shouldn't we consider the `emission_rate_per_min` and `redundancy` as parameters that are provided by the robustness layer? If we are allowing the robustness layer to configure the topology and manage that configuration, would it be better to relay all mixnet configuration aspects to the robustness layer?

I just think that it might be better to maintain all mixnet configuration logic in one place and execution in another.

mixnet/mixnet.py
```diff
@@ -1,19 +1,44 @@
...
+    __mixnode: MixNode
+
+    @classmethod
+    async def new(
+        cls,
+        initial_topology: MixnetTopology,
+        emission_rate_per_min: int,  # Poisson rate parameter: lambda
+        redundancy: int,
+        encryption_private_key: X25519PrivateKey,
+        delay_rate_per_min: int,  # Poisson rate parameter: mu
```
madxor, line 26: Also the `delay_rate_per_min` could be provided by the robustness layer.

mixnet/topology.py
```diff
@@ -0,0 +1,56 @@
...
+    def generate_route(self) -> list[MixNodeInfo]:
+        return [random.choice(layer) for layer in self.layers]
+
+    def choose_mix_destionation(self) -> MixNodeInfo:
+        all_mixnodes = [mixnode for layer in self.layers for mixnode in layer]
+        return random.choice(all_mixnodes)
+
+
+@dataclass
+class MixnetTopologySize:
```
madxor, line 31: It looks like the "MixnetTopologySize" is used mostly for the robustness layer (and test_utils), so maybe it would be better to make it part of the topology configuration inside of the robustness layer?

mixnet/client.py
youngjoon-lee: Good point. I've postponed it because in v1, only topology is updated by the robustness layer. But, yes, it's much better to move all mixnet config logics to the robustness layer, since that's why we've introduced the robustness layer. Thank you for bringing this up. I just refactored it: c402968.

mixnet/mixnet.py
youngjoon-lee: Resolved by https://github.com/logos-co/nomos-specs/pull/57#discussion_r1469009270

mixnet/topology.py
youngjoon-lee: Resolved by https://github.com/logos-co/nomos-specs/pull/57#discussion_r1469009270

al8n (approved):
👍

danielSanchezQ (approved):
Big changes, it is better to review this as a whole chunk in the branch. Looking really good. Thanks!

mixnet/client.py
```diff
@@ -1,76 +1,119 @@
...
+        if not self.__real_packet_queue.empty():
+            addr, packet = self.__real_packet_queue.get_nowait()
+            # Schedule redundant real packets
+            for _ in range(redundancy - 1):
+                redundant_real_packet_queue.put_nowait((addr, packet))
+                await self.__outbound_socket.put((addr, packet))
+
+        packet, route = PacketBuilder.drop_cover(
+            b"drop cover", self.__config.topology
+        ).next()
```
danielSanchezQ, line 106: I think the packet builder is a bit complicated to follow. Why in every instance we set the `self.iter` but is a Class attribute? Then it is wrapped in an iterator that just consumes a single packet/route?

mixnet/robustness.py
```diff
@@ -27,37 +24,77 @@ class Robustness:
...
+        for layer_id in range(mixnet_topology_size.num_layers):
+            start = layer_id * mixnet_topology_size.num_mixnodes_per_layer
+            layer = sampled[start : start + mixnet_topology_size.num_mixnodes_per_layer]
             layers.append(layer)
         return MixnetTopology(layers)
+
+
+@dataclass
+class RobustnessConfig:
+    """In v1, the robustness layer manages configs only for the mixnet layer."""
```
danielSanchezQ, line 70: We have discussed this. And I am not really happy with the naming as well. But, I think this is a coordination layer **just for the mixnet itself**. Probably for the best not to share responsibilities with any other thing.

youngjoon-lee: I've had similar thoughts, as we discussed in https://github.com/logos-co/nomos-node/pull/570#discussion_r1469895805. In my honest opinion, if this is only for mixnet, we don't really need to define this as a separate component (layer, class, service, or whatever). If it's better to have this independently for later uses, we need to clarify what this component will be responsible for other components as well as mixnet. I think we need to discuss about this more with @madxor here.

mixnet/client.py
youngjoon-lee: > Then it is wrapped in an iterator that just consumes a single packet/route?

For drop cover, yes, this class (iterator) returns only a single packet. For real traffic, the number of packets depends on the length of message provided.

> Why in every instance we set the self.iter but is a Class attribute? 

I thought it would be more readable to make PacketBuilder as an interator, but maybe it's not. I can probably inherit iterator, but I guess it would be much simpler to make it just return a list. Let me try and check which would be simpler.

youngjoon-lee: I just refactored it simpler: e0d9df1e3039fa8532fc2975720b739ac676cd71, instead of using iterator.

mixnet/robustness.py
youngjoon-lee: We're investing much time for this topic but still not making a conclusion yet. I just made a separate PR to boost our conclusion. Please have a look https://github.com/logos-co/nomos-specs/pull/61

youngjoon-lee: I'm merging this PR before it's getting hard to track all changes in this PR. This discussion will be continued in #61.

PR #58, opened 2024-01-29
Add fork choice rule

zeegomo:
Add the fork choice rule as described in 'Ouroboros Genesis: Composable Proof-of-Stake Blockchains with Dynamic Availability'

Tests still missing

### DIFF
--- a/cryptarchia/cryptarchia.py
+++ b/cryptarchia/cryptarchia.py
@@ -125,7 +125,8 @@ class Follower:
 
     # Evaluate the fork choice rule and return the block header of the block that should be the head of the chain
     def fork_choice(local_chain: Chain, forks: List[Chain]) -> Chain:
-        return maxvalid_bg(local_chain, forks)
+        # TODO: define k and s
+        return maxvalid_bg(local_chain, forks, 0, 0)
 
     def tip(self) -> BlockHeader:
         return self.fork_choice()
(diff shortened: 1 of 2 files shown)


cryptarchia/cryptarchia.py
```diff
@@ -218,5 +218,35 @@ def propose_block(self, slot: Slot, parent: BlockHeader) -> BlockHeader:
...
+    for i, (x, y) in enumerate(zip(a.blocks, b.blocks)):
+        if x.id() != y.id():
+            return i
+    return min(len(a.blocks), len(b.blocks))
+
+
+def chain_density(chain: Chain, slot: Slot) -> int:
+    return count(block for block in chain.blocks if block.slot < slot)
+
+def maxvalid_bg(local_chain: Chain, forks: List[Chain], k: int, s: int) -> Chain:
```
danielSanchezQ, line 231: Could we add a not for what `k` and `s` are?

danielSanchezQ (approved):
🎸

cryptarchia/cryptarchia.py
```diff
@@ -218,5 +214,48 @@ def propose_block(self, slot: Slot, parent: BlockHeader) -> BlockHeader:
...
+        else:
+            # The chain is forking too much, we need to pay a bit more attention
+            # In particular, select the chain that is the densest after the fork
+            forking_slot = Slot(
+                cmax.blocks[lowest_common_ancestor].slot.absolute_slot + s
+            )
+            cmax_density = chain_density(cmax, forking_slot)
+            candidate_density = chain_density(chain, forking_slot)
+            if cmax_density < candidate_density:
+                cmax = chain
```
davidrusu, line 255: thinking out loud: If the second forking rule is so great, why do we need the first one? performance? 🤔

zeegomo: I guess 's' can't be to small to avoid tails in the distribution of slot leaders, so you would need to wait a bit every time to observe the distribution after a fork. Also, if you were to propose a block after a slot with duplicate leaders you can't use a rule that needs information about future slots

PR #59, opened 2024-01-30
Clarify Cryptarchia slot calcuation

youngjoon-lee:
Please correct me if I'm wrong.

It seems that the `absolute_slot` needs to be calculated based on the `chain_start_time`, if I understand correctly. 

And, according the code here, it looks like the unit of the parameter `s` is "slots", not seconds. I added it in the comment. Did I get it correctly?


zeegomo (approved):
Good catch!

PR #60, opened 2024-01-30
Add eth-specs submodule

danielSanchezQ:
Ethereum specs have a big chunk of of primitives that we would need for building our DA specs. This will save us the time of implementing (and testing/trusting) everything ourselves. We can just focus on our higher level protocol.


youngjoon-lee (approved):
Cool!

megonen:
Great job!

PR #62, opened 2024-02-01
Follower maintains ledger state as it follows the blockchain

davidrusu:
Leader coins are now spent when they become slot leader.

Next step is to have the leaderproof produce a new coin that the validator can then use lead another slot.


cryptarchia/cryptarchia.py
```diff
@@ -93,20 +143,57 @@ def block_position(self, block: BlockHeader) -> int:
...
-    # We don't do any validation in the current version
-    def validate_header(block: BlockHeader) -> bool:
-        return True
+        self.local_chain = Chain([])
+        self.epoch = EpochState(
+            stake_distribution_snapshot=genesis_state,
+            nonce_snapshot=genesis_state,
+        )
+        self.ledger_state_snapshot = genesis_state
+        self.ledger_state = genesis_state
```
davidrusu, line 180: Note, with this change there is no genesis-block, instead we have a genesis ledger-state, and the chain is empty.

cryptarchia/cryptarchia.py
```diff
@@ -37,9 +43,48 @@ def epoch(self, config: TimeConfig) -> Epoch:
...
+
+    @staticmethod
+    def from_coin(coin: Coin):
+        return MockLeaderProof(commitment=coin.commitment(), nullifier=coin.nullifier())
+
+    def verify(self, slot):
+        # TODO: verification not implemented
+        return True
+
+    def _id_update(self, hasher):
```
zeegomo, line 85: what's this for?

cryptarchia/cryptarchia.py
```diff
@@ -93,20 +143,57 @@ def block_position(self, block: BlockHeader) -> int:
...
 
+@dataclass
+class LedgerState:
+    """
+    A snapshot of the ledger state up to some block
+    """
+
+    block: Id = None
+    nonce: Id = None
+    total_stake: int = None
```
zeegomo, line 154: isn't this a fixed value?

cryptarchia/cryptarchia.py
```diff
@@ -93,20 +143,57 @@ def block_position(self, block: BlockHeader) -> int:
...
+    def verify_committed(self, commitment: Id) -> bool:
+        return commitment in self.commitments
+
+    def verify_unspent(self, nullifier: Id) -> bool:
+        return nullifier not in self.nullifiers
+
+    def apply(self, block: BlockHeader):
+        assert block.parent == self.block
+        self.block = block.id()
+        self.nullifiers.add(block.leader_proof.nullifier)
```
zeegomo, line 167: the new commitment will be added in the next PR right?

cryptarchia/cryptarchia.py
zeegomo: It's generally useful to have a genesis block that generates a ledger state, so that you can pass it around and blocks can always assume to have a parent

cryptarchia/cryptarchia.py
```diff
@@ -131,40 +218,47 @@ def on_block(self, block: BlockHeader):
...
+
+        if new_chain == self.local_chain:
+            # we have not re-org'd therefore we can simply update our ledger state
+            # if this block extend our local chain
+            if self.local_chain.tip() == block:
+                self.ledger_state.apply(block)
+        else:
+            # we have re-org'd, therefore we must roll back out ledger state and
+            # re-apply blocks from the new chain
+            ledger_state = self.ledger_state_snapshot.copy()
```
zeegomo, line 244: nit: it's not clear that `ledger_state_snapshot` corresponds to the genesis state

Also, we might keep a ledger state for each chain to avoid this kind of re-org stuff. A linear history without rollbacks is easier to understand, when changing fork you simply switch to the ledger state of that fork

cryptarchia/cryptarchia.py
davidrusu: It's for the LeaderProof to contribute to the block id without explicitly speccing out this out right now. the details of what a leadership proof will be will depend on the cryptrography we use to implement the leader proof.

cryptarchia/cryptarchia.py
davidrusu: it (will be) growing with each block, i.e. when we apply a new block to the ledger state, we will be increasing the total stake by the block reward

cryptarchia/cryptarchia.py
davidrusu: yes, that's correct, leader proofs will generate a new commitment

cryptarchia/cryptarchia.py
davidrusu: ledger_state_snapshot does not necessarily correspond with genesis state. It is the most recently finalized ledger_state.

every 3k/f slots, it's safe to take a snapshot because we are very confident that there won't be a fork that lasts longer than 3k/f slots long.

So really the full spec will have something like:

- ledger_state_finalized
- ledger_state_finalizing
- ledger_state_current

where when `ledger_state_finalizing` is older than 3k/f slots, it transitions to `ledger_state_finalized` and `ledger_state_current`is moved to `ledger_state_finalizing`

cryptarchia/cryptarchia.py
davidrusu: I'm not convinced of that, it seems like the useful thing is knowing that every block has a "context" or "environment" (or state) that it is being evaluate against, therefore an original "context" makes sense to me.

Do you want to go with this approach until we see a case where a genesis block is useful?

zeegomo: Yes, that's ok for me

cryptarchia/cryptarchia.py
davidrusu: yeah that makes sense, I'll try it with a ledger state per fork and see how that feels

cryptarchia/cryptarchia.py
danielSanchezQ: > I'm not convinced of that, it seems like the useful thing is knowing that every block has a "context" or "environment" (or state) that it is being evaluate against, therefore an original "context" makes sense to me.
> 

This is interesting, but, could you specify what would this `state` actually is compound of? I cannot get the full idea in my head 😅

cryptarchia/messages.abnf
```diff
@@ -3,13 +3,14 @@
...
 VERSION          = %x01
-HEADER           = VERSION HEADER-UNSIGNED
+HEADER           = VERSION HEADER-UNSIGNED LEADER-PROOF
 HEADER-UNSIGNED  = %x00 HEADER-COMMON
 HEADER-COMMON    = CONTENT-SIZE CONTENT-ID BLOCK-DATE PARENT-ID
 CONTENT-SIZE     = U32
 BLOCK-DATE       = BLOCK-SLOT
 BLOCK-SLOT       = U64
 PARENT-ID        = HEADER-ID
+LEADER-PROOF     = <NOT SPECIFIED YET>
```
zeegomo, line 13: mmm, isn't the point of this to specify things? 
If we don't plan on having this ready for v1 I would remove this, as otherwise it's not a spec.

Apart from this, `HEADER-UNSIGNED` was meant to be one possible variant without leader proof. If we want to add a variant with it, I would suggest we create another rule and add it as a choice, e.g.:

```
HEADER = VERSION (HEADER-UNSIGNED / HEADER-SIGNEDv1) ; either HEADER-UNSIGNED or HEADER-SIGNEDv1 can follow 
HEADER-SIGNEDv1 = %x01 HEADER-COMMON LEADER-PROOF
```

davidrusu: I thought the intention was to have HEADER-UNSIGNED be the data that is signed over and the the "sig" goes outside of that and signs over the "unsigned" data.

> mmm, isn't the point of this to specify things?

The leader proof details depend on the proof system we will use, I think there's still value in setting up the structure that will be filled in by the CL details, no? right now the leader proof is just a mock structure that skips over the cryptography bits.

cryptarchia/cryptarchia.py
davidrusu: (discused offline, we'll drop the "snapshot" here and instead replay from genesis in the case of re-orgs)

cryptarchia/cryptarchia.py
davidrusu: for now the genesis state holds
- initial coin distribution
- initial nonce for epoch randomness

Hopefully not much more than that in the final design.

cryptarchia/messages.abnf
zeegomo: > The leader proof details depend on the proof system we will use, I think there's still value in setting up the structure that will be filled in by the CL details, no? right now the leader proof is just a mock structure that skips over the cryptography bits.

I would leave this out of the specs for v1 if the spec are meant to be a reference to implement the protocol in a compatible way. What would you do in the implementation with a spec that says "I don't know"? IMO, if we don't know how the transaction bits will look like, we better leave this off for now, but do that in a way that 1) signals it will be introduced in the future 2) sets a path for said future improvement.

The header unsigned variant was meant exactly for those two purposes, and does not need to stay in all future revisions (even though a genesis block, if we will have one, won't contain a leadership proof)

davidrusu: So for the purposes of the v1 spec, we do want mock leader proofs in the headers since these leader proofs are used in the state update logic and this logic goes beyond merely proof verification. (i.e. there is leader proof importing for orphan blocks that needs to be specified)

I could specify a `MOCK-LEADER-PROOF` here that is implementable, in fact I'll do that now, and similarly to the unsigned variant, we can replace it with the real deal once we understand CL more.

zeegomo (approved):
🚀

PR #63, opened 2024-02-05
Add epoch transition to spec

zeegomo:


cryptarchia/cryptarchia.py
```diff
@@ -36,6 +38,7 @@ def s(self):
 
 # An absolute unique indentifier of a slot, counting incrementally from 0
 @dataclass
+@functools.total_ordering
```
davidrusu, line 41: cool, didn't know about this

cryptarchia/cryptarchia.py
```diff
@@ -250,23 +274,17 @@ def on_block(self, block: BlockHeader):
...
-            # re-apply blocks from the new chain
-            ledger_state = self.genesis_state.copy()
-            for block in new_chain.blocks:
-                ledger_state.apply(block)
-
-            self.ledger_state = ledger_state
-            self.local_chain = new_chain
+        new_state = self.ledger_state[block.parent].copy()
+        new_state.apply(block)
+        self.ledger_state[block.id()] = new_state
```
davidrusu, line 287: yes, that's simpler

cryptarchia/cryptarchia.py
```diff
@@ -27,24 +28,39 @@ class TimeConfig:
 class Config:
     k: int
     active_slot_coeff: float  # 'f', the rate of occupied slots
+    # The numerator multiplier in the formula used for calculating the number of slots in an epoch.
+    # An epoch is exactly int(floor(epoch_length_multiplier * k / f slots)) long.
+    epoch_length_multiplier: int
```
davidrusu, line 33: epoch length is derived from the stabilization periods of the different snapshots taken within an epoch.

Thus, the target is  (4 + 3 + 3) * k / f
where 4 is for stabilizing the stake freezing snapshot, 3 is the buffer between frozen stake and nonce snapshot, and the last 3 is for stabilizing the nonce snapshot.

how about we decompose this multiple into those three periods?

```python
    epoch_period_stake_distribution: int
    epoch_period_nonce_buffer: int
    epoch_period_nonce: int

   @property
    def base_period_length(self) -> int:
        return int(floor(self.k / self.active_slot_coeff))

   @property
    def epoch_length(self) -> int:
        return (self.epoch_period_stake_distribution_stabilization
            + self.epoch_period_nonce_buffer
            + self.epoch_period_nonce_stabilization) * self.epoch_period
```

Then we can use these period configs when creating snapshots

cryptarchia/cryptarchia.py
```diff
@@ -167,8 +183,8 @@ def copy(self):
             block=self.block,
             nonce=self.nonce,
             total_stake=self.total_stake,
-            commitments=self.commitments.copy(),
-            nullifiers=self.nullifiers.copy(),
+            commitments=deepcopy(self.commitments),
+            nullifiers=deepcopy(self.nullifiers),
```
davidrusu, line 187: these commitments/nullifiers are never changed no? I didn't think there was a risk to a shallow copy.

But better to be paranoid 👍

zeegomo: We add the nullifier of the new leader proof, so yes that changes

cryptarchia/cryptarchia.py
zeegomo: Makes sense, and makes it impossible to configure values that do not make sense (e.g. seting `epoch_length_multiplier` to say 2 when the nonce snapshot is taken after 7k/f)

cryptarchia/cryptarchia.py
```diff
@@ -250,57 +290,60 @@ def on_block(self, block: BlockHeader):
...
+                return self.ledger_state[block.id()]
 
-@dataclass
-class EpochState:
-    # for details of snapshot schedule please see:
-    # https://github.com/IntersectMBO/ouroboros-consensus/blob/fe245ac1d8dbfb563ede2fdb6585055e12ce9738/docs/website/contents/for-developers/Glossary.md#epoch-structure
-
-    # The stake distribution snapshot is taken at the beginning of the previous epoch
-    stake_distribution_snapshot: LedgerState
+        return self.genesis_state
```
davidrusu, line 326: What does "valid" mean here? how about just `state_at_slot(...)`?

cryptarchia/cryptarchia.py
```diff
@@ -250,57 +290,60 @@ def on_block(self, block: BlockHeader):
...
+        stake_snapshot_slot = Slot((epoch.epoch - 1) * self.config.epoch_length())
+        stake_distribution_snapshot = self.get_last_valid_state(
+            chain, stake_snapshot_slot
+        )
+        # nonce snapshot is taken 7k/f slots into the previous epoch
+        # i.e. for epoch e, the snapshot is taken at the block at slot floor(7k/f) of epoch e-1
+        nonce_slot = Slot(
+            int(floor(7 * self.config.k / self.config.active_slot_coeff))
+            + stake_snapshot_slot.absolute_slot
+        )
```
davidrusu, line 340: yeah I think putting the epoch periods in the config would make this code clearer

davidrusu (approved):
looks good

PR #66, opened 2024-02-07
QUESTION: Use `LeaderProof.commitment` for `LedgerState.commitments_spend`

youngjoon-lee:
This is just a question to understand the Cryptarchia spec correctly. This might be a dumb question since my understanding of Crypsinous is low. 

It seems that `block.leader_proof.evolved_commitment` is added to both `LedgerState.commitments_spend` and `LedgerState.commitments_lead` list. I would like to ask if this is the intended behavior, just to check my understanding. If so, those two lists are going to always the same, then I'm curious why we need to maintain both. I went through the entire spec but couldn't find the reason why we need to maintain both. 
It seems that the `commitments_spend` is used to check if a coin is eligible to lead based on the epoch state, and the `commitments_lead` is used to check the same thing based on the ledger state of the parent block. Then, I guess two lists have the same role.

Or, I'd like to ask whether the `block.leader_proof.commitment` (instead of `evolved_commitment`) should be added to the `LedgerState.commitments_spend`, like a change in this PR.

Please correct me if I'm wrong. Thank you!

### DIFF
--- a/cryptarchia/cryptarchia.py
+++ b/cryptarchia/cryptarchia.py
@@ -260,7 +260,7 @@ class LedgerState:
         self.nonce = h.digest()
         self.block = block.id()
         self.nullifiers.add(block.leader_proof.nullifier)
-        self.commitments_spend.add(block.leader_proof.evolved_commitment)
+        self.commitments_spend.add(block.leader_proof.commitment)
         self.commitments_lead.add(block.leader_proof.evolved_commitment)
 
 
@@ -281,7 +281,7 @@ class EpochState:
         #
         # This verification is checking that first condition.
         #
-        # NOTE: `ledger_state.commitments_spend` is a super-set of `ledger_state.commitments_lead`
+        # NOTE: `ledger_state.commitments_spend` is a subset of `ledger_state.commitments_lead`
 
         return self.stake_distribution_snapshot.verify_eligible_to_spend(commitment)
 


youngjoon-lee:
Thank you for your great detailed explanation! I understood :+1: 

I misunderstood that the `commitments_spend` means "coins (commitments) that have been spent". But, it was actually "coins that can be spent", which also means that the coin is eligible for leadership. I now understood that this list is included in the stake distribution snapshot and is used to check whether a certain coin is old enough to be used for leadership. Therefore, I understood why 'evolved_commitment` (instead of `commitment`) should be added to the `commitments_spend` because the `evolved_commitment` is something like a coin identifier that will be used in the future. 

Also, I now understand why `commitments_lead` is necessary and why it's a subset of `commitments_spend` (It's because the `evolved_commitment` of transferred coins would be added to only `commitments_spend`). 

Thank you very much! I'm closing this PR.

zeegomo:
> Therefore, I understood why 'evolved_commitment(instead ofcommitment) should be added to the commitments_spendbecause theevolved_commitment` is something like a coin identifier that will be used in the future.

To expand a bit on this, the commitment in the proof should actually be a proof that the coin that won the election was in the set of committed coins (and such coin is consumed by the proof by revealing the nullifier) while the evolved commitment is a new commitment that will be added to the ledger

youngjoon-lee:
> To expand a bit on this, the commitment in the proof should actually be a proof that the coin that won the election was in the set of committed coins (and such coin is consumed by the proof by revealing the nullifier) while the evolved commitment is a new commitment that will be added to the ledger

Yeah, exactly. So clear. Thank you!

PR #70, opened 2024-02-08
Da specs

danielSanchezQ:
This is the main branch for nomos da specs. Currently there is a broad skeleton of what we intend to do without any of the kzg+rs implementation. This may be a bit too big already. I intend to at least fill a few more tests here and then use other branches to merge on top of this.
Notice that this is incomplete and open as a draft, but reviews are more than welcome.

### DIFF
--- a/da/dispersal.py
+++ b/da/dispersal.py
@@ -1,9 +1,8 @@
 from typing import List, Optional, Generator
 
-from da.common import Certificate
+from da.common import Certificate, NodeId
 from da.encoder import EncodedData
 from da.verifier import DABlob, Attestation
-from mixnet.node import NodeId
 
 
 class Dispersal:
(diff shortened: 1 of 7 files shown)


megonen:
I finished the first review. looks nice 🚀

bacv (approved):
Well structured, 🚀

da/dispersal.py
```diff
@@ -0,0 +1,50 @@
...
+    def _prepare_data(self, encoded_data: EncodedData) -> Generator[DABlob, None, None]:
+        assert len(encoded_data.row_commitments) == len(self.nodes)
+        assert len(encoded_data.row_proofs) == len(self.nodes)
+        columns = encoded_data.extended_matrix.columns()
+        column_commitments = encoded_data.column_commitments
+        row_commitments = encoded_data.row_commitments
+        rows_proofs = encoded_data.row_proofs
+        aggregated_column_commitment = encoded_data.aggregated_column_commitment
+        aggregated_column_proof = encoded_data.aggregated_column_proof
+        for index, (column, column_commitment, row_proofs) in enumerate(zip(columns, column_commitments, rows_proofs)):
```
youngjoon-lee, line 22: nitpick: It seems that we also assert `len(columns) == len(column_commitments) == len(self.nodes)` (or `== len(row_proofs)`, so we can make sure that the `zip` in the line 22 produces the expected number of elements. I came up with this point when reading the `disperse` function below. But, this is not very important. Also, it's okay if the definition of `columns` and `column_commitments` already infer this property.

danielSanchezQ: Good catch. I'll add this in subsequent PRs. Thanks!!

youngjoon-lee (commented):
LGTM

PR #72, opened 2024-02-14
Initial DA API Spec structure

bacv:
The idea is to compose 3 entities (Sender, Receiver, Producer) using DA executable spec components and spawn them in tests.
  - Zone - EZ/Client that 1) creates data; 2) disperses it; 3) sends to Producer.
  - Node - DA Node that receives chunks from Senders.
  - Producer - Block Producer that receives Certificates with Metadata from Senders.


da/mock/zone.py
```diff
@@ -0,0 +1,78 @@
...
+from da.mock.transport import Transport
+
+@dataclass
+class MockZoneParams:
+    da_nodes: List[MockDaNode]
+    block_producer: MockProducerNode
+    threshold: int
+    encoder_params: DAEncoderParams
+
+class MockZoneNode:
```
danielSanchezQ, line 21: We can call this `Client` or `DaClient`. Is the better abstraction as not necessarily a *Zone* is gonna be the only thing to use this.

da/mock/transport.py
```diff
@@ -0,0 +1,30 @@
...
+        receiver.receive_message(message)
+
+    def receive_message(self, message):
+        self.inbound_messages.append(message)
+        if self.message_handler:
+            self.message_handler(self, message)
+
+    def connect(self, other_node):
+        self.neighbors[other_node.identifier] = other_node
+
```
danielSanchezQ, line 30: We can probably get rid of this as it is not necessary for the purpose of this specs.

da/mock/common.py
```diff
@@ -0,0 +1,43 @@
...
+
+@dataclass
+class DABlobMessage:
+    blob: DABlob
+    metadata: Metadata
+
+@dataclass
+class CertificateMessage:
+    certificate: Certificate
+    metadata: Metadata
```
danielSanchezQ, line 28: I think this is the most important part. A storage/retrieval specification on how to store and query this data is what I feel is missing.

danielSanchezQ (commented):
I added some comments to have a better context. As we are gonna work on integrating stuff this will be done in further PRs that we can merge on top of this one.

danielSanchezQ (approved):

PR #73, opened 2024-02-26
KZG core functionality

danielSanchezQ:
This is the easiest workig version of kzg we could came up with.
Commitment + Proof + Verification
Almost everything implemented from scratch for easier understanding.
No reed solomon involved which will be introduced later on in a different PR

### DIFF
--- a/da/test_kzg.py
+++ b/da/test_kzg.py
@@ -29,3 +29,4 @@ class TestKZG(TestCase):
         commit = kzg.bytes_to_commitment(rand_bytes, GLOBAL_PARAMETERS)
         poly = kzg.bytes_to_polynomial(rand_bytes)
         proof = kzg.generate_element_proof(poly[0], poly, GLOBAL_PARAMETERS)
+        self.assertEqual(len(proof), 48)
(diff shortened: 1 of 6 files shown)


megonen:
Maybe adding "verify setup" can be helpful to check the integrity of trusted setup procedure.

megonen:
LGTM

da/test_kzg.py
```diff
@@ -0,0 +1,61 @@
...
+
+from da.kzg_rs import kzg
+from da.kzg_rs.common import BYTES_PER_FIELD_ELEMENT, GLOBAL_PARAMETERS, ROOTS_OF_UNITY, GLOBAL_PARAMETERS_G2
+from da.kzg_rs.trusted_setup import verify_setup
+
+
+class TestKZG(TestCase):
+
+    @staticmethod
+    def rand_bytes(size=1024):
```
youngjoon-lee, line 15: I'm late to review this PR, but have a quick question. It seems like this parameter `size` should be `n_chunks` if I understand correctly?

danielSanchezQ: yes it is

PR #76, opened 2024-03-05
Da encoding protocol

danielSanchezQ:
* Fill up the encoding protocol structure with the actual specification
* Expand an implement tests


megonen:
I have a new idea. We can publish column commitments, but when calculating the agg_com value, we can calculate the commitment value of H(ColumnComm)_i values. This way, if someone wants to add an extra column, they will not be able to add it because it is over the agg_com hash, and we will provide the things we provided in our previous design.

megonen (approved):
LGTM

PR #78, opened 2024-03-08
Da verifier protocol

danielSanchezQ:
Fill up verifier and verifier tests
This is lacking the attestation creation which is not yet defined in the specification. It can be added later in a different PR.


megonen (approved):
done 👍

PR #81, opened 2024-03-13
Threshold bitfield on certificate

danielSanchezQ:


da/dispersal.py
```diff
@@ -69,9 +76,18 @@ def _build_attestation_message(encoded_data: EncodedData) -> bytes:
...
         attestations = []
         attested_message = self._build_attestation_message(encoded_data)
-        for node, pk, blob in zip(self.settings.nodes_ids, self.settings.nodes_pubkey,  self._prepare_data(encoded_data)):
+        signed = Bitfield(False for _ in range(len(self.settings.nodes_ids)))
+        blob_data = zip(
+            range(len(self.settings.nodes_ids)),
+            self.settings.nodes_ids,
+            self.settings.nodes_pubkey,
+            self._prepare_data(encoded_data)
+        )
```
davidrusu, line 85: Do you think it's worth sorting `settings.nodes_ids` so that we are sure that everyone is mapping each node id to the same index.

da/dispersal.py
```diff
@@ -45,11 +45,18 @@ def _prepare_data(self, encoded_data: EncodedData) -> Generator[DABlob, None, No
...
+    ) -> Certificate:
         assert len(attestations) >= self.settings.threshold
+        assert len(attestations) == signers.count(True)
         aggregated = bls_pop.Aggregate([attestation.signature for attestation in attestations])
         return Certificate(
             aggregated_signatures=aggregated,
+            signers=signers,
             aggregated_column_commitment=encoded_data.aggregated_column_commitment,
             row_commitments=encoded_data.row_commitments
         )
```
davidrusu, line 62: Are these certificates verified anywhere? I don't see it in this PR

da/dispersal.py
danielSanchezQ: Yeah, that could be useful. But I have been thinking actually it may go further than that:
My initial idea was that ordering/participation_set was coming from the on-chain snapshot of some specific epoch. But I am not really sure it can be done like that, or if it actually should. Otherwise, participants that attested DA may be gone by the time the data is included in a block. In that case we would need some extra validation process (?)
@megonen @alvatar 

PD: I'll order it for now.

davidrusu (approved):

PR #82, opened 2024-03-14
Added verification to certificate

danielSanchezQ:


da/common.py
```diff
@@ -55,3 +55,16 @@ class Certificate:
     aggregated_column_commitment: Commitment
     row_commitments: List[Commitment]
 
+    def verify(self, nodes_public_keys: List[BLSPublickey]) -> bool:
+        # we short them as the signers bitfield is sorted by the public keys as well
```
bacv, line 59:
```suggestion
        # we sort them as the signers bitfield is sorted by the public keys as well
```

bacv (approved):

megonen (commented):
to prevent Rogue Key Attack mentioned here https://www.notion.so/BLS-Aggregated-and-Threshold-Signature-5a37ed1db01e4f9594080e1e6a5ef5f4, test can be added for PoP (proof of possesion) mechanism

da/common.py
```diff
@@ -37,21 +37,43 @@ def transposed(self) -> Self:
...
     pass
 
 
 @dataclass
 class Attestation:
-    signature: BLSSignature
+    # Signature over the attestation message H(AggC, RowC...)
+    message_signature: BLSSignature
+    # Proof of possession on bls pop sk
+    proof_of_possession: BLSSignature
```
davidrusu, line 52: @megonen Proof of Possession should go with the registration of your public key, right? We don't need to proof possession with each attestation?

mentioning because I believe putting PoP in attestation/certificate will blow up the certificate size

danielSanchezQ: So, thing goes like this:
* Upon registering the pk nodes also need to provide the pop 
* We keep them *both* available
* pop is only verified upon registration, meaning that we can really consider the pks trustful here already?

danielSanchezQ: I'll revert the latest commits then and just add an the pop in the initialization as a constrain.

da/common.py
```diff
@@ -55,3 +55,21 @@ class Certificate:
...
+    def verify(self, nodes_public_keys: List[BLSPublickey]) -> bool:
+        """
+        List of nodes public keys should be a trusted list of verified proof of possession keys.
+        Otherwise, we could fall under the Rogue Key Attack
+        `assert all(bls_pop.PopVerify(pk, proof) for pk, proof in zip(node_public_keys, pops))`
+        """
+        # we sort them as the signers bitfield is sorted by the public keys as well
+        signers_keys = list(compress(sorted(nodes_public_keys), self.signers))
+        message = build_attestation_message(self.aggregated_column_commitment, self.row_commitments)
+        return bls_pop.AggregateVerify(signers_keys, [message]*len(signers_keys), self.aggregated_signatures)
```
danielSanchezQ, line 67: @davidrusu I added this comment. This keys verification should be done upon registration.
@megonen We have to update the specification documents with this most probably.

davidrusu (approved):
🚀

PR #84, opened 2024-03-18
cryptarchia: fix try_create_fork to find parent block

youngjoon-lee:
It seems that we should find the position of the "parent" of the block to be added when trying to create a fork, but the current implementation finds the position of the "block" to be added.

Please correct me if I'm wrong.

### DIFF
--- a/cryptarchia/cryptarchia.py
+++ b/cryptarchia/cryptarchia.py
@@ -244,14 +244,11 @@ class Chain:
     def length(self) -> int:
         return len(self.blocks)
 
-    def contains_block(self, block: BlockHeader) -> bool:
-        return block in self.blocks
-
-    def block_position(self, block: BlockHeader) -> int:
-        assert self.contains_block(block)
+    def block_position(self, block: Id) -> int:
         for i, b in enumerate(self.blocks):
-            if b == block:
+            if b.id() == block:
                 return i
+        return -1
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


youngjoon-lee:
> Do you want to add a test to verify that forks are created correctly?

Sure. I'll add tests in this PR soon and re-request your review!

zeegomo:
Good catch, concerning that it was not caught by tests

cryptarchia/cryptarchia.py
```diff
@@ -244,14 +244,11 @@ def tip(self) -> BlockHeader:
...
-        return block in self.blocks
-
-    def block_position(self, block: BlockHeader) -> int:
-        assert self.contains_block(block)
+    def block_position(self, block: Id) -> int:
         for i, b in enumerate(self.blocks):
-            if b == block:
+            if b.id() == block:
                 return i
+        return -1
```
danielSanchezQ, line 251: nitpick: I would rather return an `Optional[int]` instead for the `-1` in case is not found.

youngjoon-lee: yeah, thank you.0722830bfccaea3319504770bd9d1aceb56e2e63

youngjoon-lee: Yeah. Thank you f35e7ba

cryptarchia/cryptarchia.py
```diff
@@ -420,10 +417,10 @@ def try_create_fork(self, block: BlockHeader) -> Optional[Chain]:
 
         chains = self.forks + [self.local_chain]
         for chain in chains:
-            if chain.contains_block(block):
-                block_position = chain.block_position(block)
+            block_position = chain.block_position(block.parent)
+            if block_position != -1:
                 return Chain(
-                    blocks=chain.blocks[:block_position],
+                    blocks=chain.blocks[: block_position + 1],
```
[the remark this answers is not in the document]
youngjoon-lee: @davidrusu @zeegomo I just added a test that covers all cases of fork creation. 2460827

Previously, this case hasn't been caught by tests because we've only tested the case which a fork is created from the genesis block, not from the middle of the chain. https://github.com/logos-co/nomos-specs/blob/b1e13f79c5e4ddefbd8424a3188bc093c355d234/cryptarchia/test_ledger_state_update.py#L44

davidrusu (approved):

PR #85, opened 2024-03-18
Da api ext tests

bacv:
Couple tests to demonstrate storing of the same blob with different indexes.

### DIFF
--- a/da/api/test_flow.py
+++ b/da/api/test_flow.py
@@ -22,14 +22,14 @@ class MockStore(BlobStore):
         if metadata.index in self.app_id_store[metadata.app_id]:
             raise ValueError("index already written")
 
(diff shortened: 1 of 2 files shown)


danielSanchezQ:
@davidrusu @megonen @alvatar 
We realized that publishing exactly the same data could happen. At the moment the identification for a published blob depend on the data itself, so publishing the same blob twice (or more) would result in the same identification. This opens a few questions:

* I think is an opportunity for optimization (?) you do not need to store a blob if you already have it, you just add an index to the same data (?)
* Should we rather consider them independent so they are duplicated? (This would result in adding an specific information to the identifier hash).

danielSanchezQ (approved):

PR #88, opened 2024-03-20
DA: Verifier skip duplicated blobs intents

danielSanchezQ:
Verifiers should know what was already attested, and if it was already attested just skip next same blobs verification attempts.

### DIFF
--- a/da/test_verifier.py
+++ b/da/test_verifier.py
@@ -26,14 +26,12 @@ class TestVerifier(TestCase):
             )
         )
 
-    def test_build_attestation(self):
-        pass
-
     def test_verify(self):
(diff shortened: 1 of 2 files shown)


bacv (approved):

megonen (approved):
Is there a possibility of this happening? I mean, why should the previously attested blob  come back for verification again?

bacv (approved):

danielSanchezQ:
> Is there a possibility of this happening? I mean, why should the previously attested blob come back for verification again?

As the protocol is trustless we do not know why or who may request it, but we need to know how to react to that situations anyway.

PR #93, opened 2024-05-20
WIP: Cl/executable spec

davidrusu:


coordination-layer/crypto.py
```diff
@@ -0,0 +1,108 @@
...
+    # point. We will seed the random module with the result of hashing the
+    # elements and then call Point.random() to retreive the point
+    # corresponding to the mentioned elements.
+
+    r = prf(f"HASH_TO_CURVE_{domain}", *elements)
+
+    import random
+
+    random.seed(r.v)
+    return Point.random()
```
danielSanchezQ, line 72: Not sure, but probably better to reset the seed to the original state. Or it may affect other computations based on calling this function.

davidrusu: true, let me see if I can retrieve the current global random state somehow.

danielSanchezQ: > true, let me see if I can retrieve the current global random state somehow.

I edited the comment and added some reusable method.

davidrusu: Interesting, I was going to use the `getstate()/setstate()` api: https://docs.python.org/3/library/random.html#random.getstate

```
    old_state = random.getstate()

    random.seed(r.v)
    r = Point.random()

    random.setstate(old_state())
    return r
```

PR #94, opened 2024-05-23
DA domain specific tag for bls methods

bacv:
Added domain specific tag for signing and verifying Nomos DA related data.


megonen (approved):
LGTM 🚀

PR #98, opened 2024-06-26
WIP: Mixnet: Update spec for NomMix

youngjoon-lee:
Reference: [NomMix: Peer-to-Peer Mix Network over Gossip Channel](https://www.notion.so/NomMix-Peer-to-Peer-Mix-Network-over-Gossip-Channel-11e9c70a656d4ff1858bc46a715f4c59?pvs=4)

Now, the spec is much simpler.
- Deleted many files not necessary anymore
- Added the "Global Transmission Rate" (aka. GTR)
- Added the per-conn 0-hop noise
- No mixnode topology anymore.
  - Now we selects k nodes randomly from the membership table.
  - How the membership table will be disseminated is not determined yet. Maybe it will be gossiped through the network. Here, we assume that it is already disseminated completely to each node.
- No direct connection between mix nodes anymore.
  - All messages are gossiped. Only one node in the network can unwrap the Sphinx packet if he has a correct private key.

This spec will be used for the NomMix simulation. I'm already working on it. 

TODO (Written as TODO commented in the source code):
- Temporal mixing
- Limiting the max length of mix route to make the size of Sphinx packet constant.


mixnet/config.py
```diff
@@ -2,110 +2,56 @@
...
+    node_configs: List[NodeConfig]
+    membership: MixMembership
 
 
 @dataclass
-class MixnetTopologyConfig:
-    mixnode_candidates: List[MixNodeInfo]
-    size: MixnetTopologySize
-    entropy: bytes
+class NodeConfig:
```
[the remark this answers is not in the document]
youngjoon-lee: I think the max number of connections is the part of underlying gossip protocol but should be set when running the node. The preferred and maximum length of k-path should be a node parameter. 
I think it's meaningful to specify the k-path length parameter in this spec. I'll do it as a next PR. Especially the max length is important for limiting the size of Sphinx packet as a constant regardless of the actual k-path length that each node will choose. I'll clarify it in this spec in the next PR.
Also, the number of connections is important for defining the proper GTR (Global Transmission Rate) because it's related to the max bandwidth capacity of each node. I'm not sure if we need to specify it in this spec, but I'll define/use this param for simulation.

mixnet/fisheryates.py
```diff
@@ -1,21 +0,0 @@
-import random
-from typing import List
-
-
-class FisherYates:
```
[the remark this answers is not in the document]
youngjoon-lee: Yes. I'm removing this because the queuing mechanism will be implemented in the next PR. I'll revive this in the next PR if it's needed for queuing.

mixnet/config.py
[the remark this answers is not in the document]
youngjoon-lee: This is a good point. Thank you. I added a commit 3585d3c which clarify what's a part of global config and what's a local node config.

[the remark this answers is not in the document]
madxor: I think that number of connections and GTR are important from the perspective of simulation. These two parameters (alongside the queueing mechanism) will define the speed at which messages are propagated through the network. This might not be important from the spec perspective, but if this is important from the simulation perspective then I think it should also be relevant from the specification perspective. Nevertheless, you are an expert in this field so I trust your judgement :)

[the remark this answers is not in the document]
youngjoon-lee: While working on the simulation, now I think it’s better to have a param for the num of connections :) I’ll add it in this PR soon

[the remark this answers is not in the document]
youngjoon-lee: @madxor ed9b0ab I added the `peering_degree` param to the `NodeConfig`, so that it can be configured by each node. From its [definition](https://docs.libp2p.io/concepts/pubsub/overview/#full-message) in libp2p gossipsub, it means the **target** number of peers that a node should maintain. For the simplicity in this spec, we use it as the **maximum** number of peers that a node can maintain because I think we don't need to stick too much to libp2p in this spec. I think the spec needs to be generic and it's enough to have only the **maximum** number of peers as a parameter in this spec.

mixnet/config.py
```diff
@@ -2,110 +2,57 @@
...
         """
         Generate a mix route for a Sphinx packet.
         The pre-selected mix_destination is used as a last mix node in the route,
         so that associated packets can be merged together into a original message.
         """
-        route = [random.choice(layer) for layer in self.layers[:-1]]
-        route.append(mix_destination)
+        route = [self.choose() for _ in range(num_hops - 1)]
+        route.append(last_mix)
         return route
```
danielSanchezQ, line 42: Nitpick:
```suggestion
        return [*(self.choose() for _ in range(num_hops - 1)), last_mix]
```

mixnet/node.py
```diff
@@ -1,107 +1,186 @@
...
+from mixnet.config import GlobalConfig, NodeConfig
+from mixnet.packet import Fragment, MessageFlag, MessageReconstructor, PacketBuilder
 
-class MixNode:
-    """
-    A class handling incoming packets with delays
+NetworkPacket: TypeAlias = "SphinxPacket | bytes"
+NetworkPacketQueue: TypeAlias = "asyncio.Queue[NetworkPacket]"
+Connection: TypeAlias = NetworkPacketQueue
+BroadcastChannel: TypeAlias = "asyncio.Queue[bytes]"
```
danielSanchezQ, line 21: Why the types are stringified (`"`)? Just curious here

mixnet/node.py
```diff
@@ -1,107 +1,186 @@
...
+        self, packet: SphinxPacket
+    ) -> NetworkPacket | None:
+        try:
+            processed = packet.process(self.config.private_key)
+            match processed:
+                case ProcessedForwardHopPacket():
+                    return processed.next_packet
+                case ProcessedFinalHopPacket():
+                    await self.__process_sphinx_payload(processed.payload)
+        except Exception:
```
danielSanchezQ, line 50: Better to be clear with exceptions. What can fail here?

mixnet/node.py
```diff
@@ -1,107 +1,186 @@
...
-            task.add_done_callback(self.tasks.discard)
+            elem = await conn.recv()
+            # In practice, data transmitted through connections is going to be always 'bytes'.
+            # But here, we use the SphinxPacket type explicitly for simplicity
+            # without implementing serde for SphinxPacket.
+            if isinstance(elem, bytes):
+                assert elem == build_noise_packet()
+                # Drop packet
+                continue
+            elif isinstance(elem, SphinxPacket):
```
danielSanchezQ, line 129: Nitpick:  You can probably [pattern-match](https://peps.python.org/pep-0636/) here

mixnet/node.py
youngjoon-lee: Removed `"`s. I added that for some reason previously but not needed anymore: 235c703

mixnet/node.py
youngjoon-lee: 235c703 I changed it to `ValueError` because `pysphinx` package doesn't raise well-defined exceptions yet. This is the case which the sphinx packet cannot be unwrapped by the private key of this node. In this case, the node should gossip it just as it is. I also updated the comment more clearly.

mixnet/config.py
youngjoon-lee: Thanks 3a70343

mixnet/node.py
youngjoon-lee: 4aa1ee0 Now that I implemented serde for SphinxPacket in `pysphinx` package, I simplified this code using pattern matching.

mixnet/config.py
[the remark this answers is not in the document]
madxor: Thanks! I agree also I think that we should treat the `peering_degree` as "expected" number of connections a node will have open. This is good enough for our simulation purposes as we can evaluate the properties of the system assuming unified level of connectivity for all nodes.

However, in real life a node might deviate from this and in current design we might not have any tool to prevent the node from opening to many or to little connections.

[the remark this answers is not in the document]
youngjoon-lee: True. In libp2p, this parameter means the 'expected' number of connections. I clarified the comment in the code: d3e8b02. This parameter will be used as the 'max' value in the simulation, for simplicity.

https://docs.libp2p.io/concepts/pubsub/overview/#full-message
> In libp2p’s default implementation the ideal network peering degree is 6 with anywhere from 4–12 being acceptable.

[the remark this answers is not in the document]
youngjoon-lee: After merging this PR to the master, I'll open subsequent PRs soon.

mixnet/node.py
```diff
@@ -1,107 +1,159 @@
...
-        cls,
-        config: MixNodeConfig,
-    ) -> Self:
-        self = cls()
+    config: NodeConfig
+    global_config: GlobalConfig
+    mixgossip_channel: GossipChannel
+    reconstructor: MessageReconstructor
+    broadcast_channel: BroadcastChannel
+    packet_size: int
```
madxor, line 35: Shouldn't we extract the `packet_size` as a global parameter? Similar to the GTR.

mixnet/node.py
```diff
@@ -1,107 +1,159 @@
...
+    async def __process_msg(self, msg: bytes) -> bytes | None:
+        """
+        A handler to process messages received via gossip channel
+        """
+        flag, msg = Node.__parse_msg(msg)
+        match flag:
+            case MsgType.NOISE:
+                # Drop noise packet
+                return None
+            case MsgType.REAL:
```
madxor, line 58: Hmm, if this is the node logic for processing messages received via gossip channel then the REAL message handler should do 2 things.
1) The handler should immediately pass the real message to the gossip routine, where it should be added to connection queues that have not seen the message.
2) Immediately after the above (or even concurrently) the handler should pass the real message to the sphinx processing routine, which after processing will add the output message to the queue or broadcast it.

In here I see that we are only processing the message in the sphinx routine but I don't see the gossip part. Can you point me to the place that we are doing the 1) part?

mixnet/node.py
```diff
@@ -1,107 +1,159 @@
...
+
+    def connect(self, peer: Node):
         """
+        Establish a duplex connection with a peer node.
+        """
+        noise_msg = Node.__build_msg(MsgType.NOISE, bytes(self.packet_size))
+        inbound_conn, outbound_conn = asyncio.Queue(), asyncio.Queue()
+
+        # Register a duplex connection for its own use
+        self.mixgossip_channel.add_conn(
```
madxor, line 103: We can use "nomssip" name for out variant of gossip protocol ;)

mixnet/config.py
```diff
@@ -2,110 +2,77 @@
...
-            layers.append(layer)
-        self.layers = layers
+    nodes: List[NodeInfo]
 
-    def generate_route(self, mix_destination: MixNodeInfo) -> list[MixNodeInfo]:
+    def generate_route(self, num_hops: int, last_mix: NodeInfo) -> list[NodeInfo]:
         """
         Generate a mix route for a Sphinx packet.
         The pre-selected mix_destination is used as a last mix node in the route,
         so that associated packets can be merged together into a original message.
```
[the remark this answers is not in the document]
youngjoon-lee: >  Or does the "merge" refer to a situation when multiple sphinx packets are needed for sending the message?

Yes. This is the implementation from the previous design where a single message is split into multiple Sphinx packets and they are merged in the last mix, if the message is big. In the current design, a single message will fit into a single Sphinx packet because the message will be small. Even if the message is not small enough, we can increase the capacity of Sphinx packet, so that the message is not fragmented. 
However, I didn't remove this fragmentation/merge feature for flexibility. If you think this makes confusion and make our design complicated, I can remove this and make sure that a message is transformed to a single Sphinx packet.

mixnet/config.py
```diff
@@ -2,110 +2,77 @@
...
-    ) -> None:
-        """
-        Build a new topology deterministically using an entropy and a given set of candidates.
-        """
-        shuffled = FisherYates.shuffle(config.mixnode_candidates, config.entropy)
-        sampled = shuffled[: config.size.num_total_mixnodes()]
+class MixMembership:
+    """
+    A list of public information of nodes in the network.
+    We assume that this list is eventually known to all nodes in the network (e.g. via p2p advertising).
```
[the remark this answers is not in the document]
youngjoon-lee: Makes sense. I'll remove the `(e.g. via p2p advertising)` and `eventually` parts because this information is probably going to be shared as an on-chain data. Since it is not determined 100% clearly, I won't mention anything specifically. Instead, just
```
We assume that this list is known to all nodes in the network.
```

mixnet/config.py
```diff
@@ -2,110 +2,77 @@
...
+    """
 
-    def encryption_public_key(self) -> X25519PublicKey:
-        return self.encryption_private_key.public_key()
+    public_key: X25519PublicKey
 
-    def sphinx_node(self) -> Node:
-        return Node(self.encryption_private_key, self.addr)
+    def sphinx_node(self) -> SphinxNode:
+        # TODO: Use a pre-signed incentive tx, instead of NodeAddress
```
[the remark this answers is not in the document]
youngjoon-lee: This is from the Alvaro's design proposal: https://www.notion.so/Maximally-decentralized-and-homogeneous-private-consensus-6df631b1c7ad430e8846f64fb874ccda?pvs=4#53ef5eb6d7ff447499e20562551f1507
> Every layer of the message has an additional field: a pre-signed transaction that can be redeemed by the receiver to incentivize relaying (ie mixing).

I added this comment before we meet in the offsite. If this is not the case in our design, of if it is not determined yet (I guess), I can remove this comment.

mixnet/node.py
youngjoon-lee: I thought about it. But, I realized that it's hard to estimate the actual packet size when configuring nodes, because the actual packet size is determined by the Sphinx packet encoding protocol, which depends on the max length of mix path (in the global config). Of course, we can provide a equation for users to calculate the actual packet size based on the max path length. But, I think it doesn't provide better user experience. Instead, I decided to calculate the actual packet size automatically when the node is initialized, so that it can be used when generating noise packet (in the same size as the real packet, which is a Sphinx).

mixnet/node.py
youngjoon-lee: Good point. The part 1) can be found [here](https://github.com/logos-co/nomos-specs/blob/871215742984bface1878b943c9d8f7506a10351/mixnet/node.py#L80-L80). The `__process_sphinx_packet` returns the original packet as it is, so that it can be gossiped through. But, I agree that it's not readable. Also, as you mentioned, the part 1) must be done by the gossip channel level, not in the node logic level. When imagining the libp2p gossipsub, the libp2p will anyway gossip all incoming messages (unless they're duplicate) regardless of the node internal logic. Also, that's why the noise packet needs to be dropped in the gossip channel level (, not the node logic level), to ensure that any noise isn't gossiped further. I accidentally moved this noise handling into the node logic when I refactored the code. I'll fix this right now and ping you again.

mixnet/node.py
youngjoon-lee: Good naming. 953b2d6 renamed it.

mixnet/node.py
youngjoon-lee: 010733c I refactored the Nomssip v.s. Node encapsulation to clarify the part 1) and 2). Now, the logic flow in the source code is very clear.
The part 1) is done in the nomssip level. The part 2) is done in the node logic level.

mixnet/config.py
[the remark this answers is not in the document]
youngjoon-lee: 2d703ca I just removed all fragment-related code. In the new NomMix spec, we're assuming that a single message is small enough to fit into a single Sphinx packet. We're designing bandwidth-related properties based on this assumption. So, it's more reasonable to not have fragment-related code in this executable spec. If the message needs to be bigger, we can modify the Sphinx spec to accept a bigger message for a single Sphinx packet, as I mentioned [above](https://github.com/logos-co/nomos-specs/pull/98#discussion_r1673247324).

mixnet/connection.py
```diff
@@ -0,0 +1,60 @@
...
+
+class MixSimplexConnection:
+    """
+    Wraps a SimplexConnection to add a transmission rate and noise to the connection.
+    """
+
+    queue: NetworkPacketQueue
+    conn: SimplexConnection
+    transmission_rate_per_sec: int
+    noise_msg: bytes
```
danielSanchezQ, line 37: Likewise, why is this declared here?

If it is for typing, you can add it it to the `self` declarations below I think 🤔

mixnet/config.py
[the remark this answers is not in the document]
madxor: Great! Thank you!

mixnet/config.py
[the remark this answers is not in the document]
madxor: Great, thank you! I understand the flexibility of the fragmentation, however, as you also note, we would need to gossip every fragment of the message. This is not a big issue but could significantly increase the message dissemination as each fragment will be independently delayed on every node, and the receiving node will need to wait for all fragments to proceed (with cryptographical mix).

mixnet/nomssip.py
```diff
@@ -0,0 +1,125 @@
...
+        # https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
+        self.tasks = set()
+
+    def add_conn(self, inbound: SimplexConnection, outbound: SimplexConnection):
+        if len(self.conns) >= self.config.peering_degree:
+            # For simplicity of the spec, reject the connection if the peering degree is reached.
+            raise ValueError("The peering degree is reached.")
+
+        noise_packet = self.__build_packet(
+            self.PacketType.NOISE, bytes(self.config.msg_size)
```
danielSanchezQ, line 48: do we send empty bytes? Wouldn't this be a pattern that could be use to denoise?

mixnet/config.py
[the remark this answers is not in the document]
madxor: Incentivization is not part of the design so it might be confusing for the reader. It would be better to remove it and introduce it again when we design that part also.

mixnet/nomssip.py
```diff
@@ -0,0 +1,125 @@
...
+        return flag.value + data
+
+    @staticmethod
+    def __parse_packet(data: bytes) -> tuple[PacketType, bytes]:
+        """
+        Parse the message and extract the flag.
+        """
+        if len(data) < 1:
+            raise ValueError("Invalid message format")
+        return (Nomssip.PacketType(data[:1]), data[1:])
```
danielSanchezQ, line 125: I think is clearer if you use indexing instead of slicing for the first element.
```suggestion
        return (Nomssip.PacketType(data[0]), data[1:])
```

mixnet/node.py
madxor: So maybe we could define global `max_message_size` which then will be used to calculate the `sphinx_packet_size` (I'm using this instead of `packet_size` as it is more precise to me)? I'm imagining that it should be something like `sphinx_packet_size = sphinx_header_size(max_mix_hops) + max_message_size`.

youngjoon-lee: The calculation of `sphinx_packet_size` is little bit more complex. So yeah, it is good to have the global parameter `max_message_size`, but I’d suggest to calculate the `sphinx_packet_size` internally by creating a sample sphinx packet instead of calculating an equation that can be probably changed by the Sphinx library. Of course, we’ll maintain the Sphinx library ourselves. We can later add a function in the library to calculate the size without creating a sample

mixnet/connection.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class DuplexConnection:
+    """
+    A duplex connection in which data can be transmitted and received simultaneously in both directions.
+    This is to mimic duplex communication in a real network (such as TCP or QUIC).
+    """
+
+    inbound: SimplexConnection
+    outbound: MixSimplexConnection
```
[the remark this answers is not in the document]
youngjoon-lee: Oh, my bad. I was confused because of dataclasses. I removed class variables: e1abc44

mixnet/connection.py
youngjoon-lee: thank you e1abc44

mixnet/node.py
madxor: Looks good. One thing to consider, have noticed that we are using `gossip` and `__gossip`, wouldn't it be more readable to use something more descriptive? Like `gossip_new` for gossiping new messages or `gossip_relay` for gossiping received messages? It's totally a nit, and I'm fine with just gossip as we have comments :)

mixnet/nomssip.py
madxor: We are assuming TLS on the connection level so for the observer it will be indistinguishable from real traffic. Peers must know that we are sending noise to let them drop that, so we can use zeros here or any known random sequence.

mixnet/config.py
```diff
@@ -2,110 +2,69 @@
...
-    num_layers: int
-    num_mixnodes_per_layer: int
-
-    def num_total_mixnodes(self) -> int:
-        return self.num_layers * self.num_mixnodes_per_layer
-
-
-# 32-byte that represents an IP address and a port of a mix node.
-NodeAddress: TypeAlias = bytes
+        return [random.choice(self.nodes) for _ in range(length)]
```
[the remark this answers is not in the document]
youngjoon-lee: I just modified this code to use `choices` which is much simpler and maybe efficient. 
e558ba4e6d735a5c388f0ad8f3b37bbd0473f8ac
But, I don't use `random.sample` here because I think we don't need to prevent the case which the same node is chosen multiple times. I updated the comment for this.

mixnet/config.py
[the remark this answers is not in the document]
youngjoon-lee: I agree. I just removed the comment: e964721

mixnet/nomssip.py
youngjoon-lee: Yes. Even if we use random sequence (instead of zeros), packets will be distinguishable if we don't have secure transport such as TLS. So here, we assume that we will always have secure transport.

mixnet/nomssip.py
youngjoon-lee: Right. Thank you.

mixnet/node.py
madxor: Yes, you are right, it's better not to design a function to calculate the `sphinx_packet_size` by hand and getting that value from sample sphinx packet.

mixnet/node.py
youngjoon-lee: I agree. I just made it more readable by making `__gossip` to accept only `FlaggedPacket` object that I just defined. (also, renamed the function)

mixnet/node.py
youngjoon-lee: c360a4a I just added a new global parameter `max_message_size`. If a node tries to send a message longer than that, a `ValueError` will occur. All messages shorter than that will be padded by the sphinx encoding to make all sphinx packets have the same size.

I just found that the current Sphinx library (that I implemented) have a hard-coded max message size (that is defined by Loopix Sphinx). I'll make it flexible, so that our `max_message_size` parameter is used for that.

youngjoon-lee:
I just opened a new PR #109 that moves the previous mixnet spec into a new `deprecated/` directory to make the repo clean. Once the PR is merged, I will rebase this PR to the new `master`, so that this PR can add new spec files to the new `mixnet/` directory. Until then, I am marking this PR as a draft.

youngjoon-lee:
Closing this, as we discussed in Discord. The executable spec needs to be written with the technical spec, after the design is verified by simulations and PoC. Although I put only clear parts of the spec in this PR, which is not likely to change during further experiments, we are not sure right now. So, let's close this PR for now, and recreate a new PR once we are ready to write the technical spec.

PR #100, opened 2024-07-08
Direct DA connection protocol

bacv:
A proposal for executor to da node direct communication protocol, intended to use in DA PoC and support protocol improvements over time.

This PR is supposed to be merged into `da-poc` branch.


danielSanchezQ:
Why is it mixed with CL stuff?

bacv:
This happened after I've changed the target branch from `master` to `da-poc`. Fixed.

da/executor/dispersal.proto
```diff
@@ -0,0 +1,60 @@
+syntax = "proto3";
+
+package dispersal;
+
+message Blob {
```
bacv, line 5: Ideally this should go into a separate file, where common DA protocol types are defined.

da/executor/executor.py
```diff
@@ -0,0 +1,49 @@
...
+        self.interval = 10
+
+    async def execute(self):
+        blob_id = b"test"
+        data = b"TEST DATA"
+        while True:
+            try:
+                for transport in self.connections:
+                    message = proto.new_dispersal_req_msg(blob_id, data)
+                    await transport.write(message)
```
danielSanchezQ, line 20: Nitpick: Did you try to write it councurrently?

da/executor/executor.py
```diff
@@ -0,0 +1,49 @@
...
+        except Exception as e:
+            print(f"Executor: Failed to connect or lost connection: {e}")
+
+    async def _handle(self, conn_id, writer, message):
+        if message.HasField('dispersal_res'):
+            print(f"Executor: Received DispersalRes: blob_id={message.dispersal_res.blob_id}")
+        elif message.HasField('sample_res'):
+            print(f"Executor: Received SampleRes: blob_id={message.sample_res.blob_id}")
+        else:
+            print("Executor: Received unknown message type")
```
danielSanchezQ, line 44: Maybe print out the message for debugging purposes?

da/executor/executor.py
bacv: You mean something like that?:
```python
    async def execute(self):
        blob_id = b"test"
        data = b"TEST DATA"
        while True:
            try:
                tasks = []
                for transport in self.connections:
                    message = proto.new_dispersal_req_msg(blob_id, data)
                    task = asyncio.create_task(transport.write(message))
                    tasks.append(task)
                await asyncio.gather(*tasks)
                await asyncio.sleep(self.interval)
            except asyncio.CancelledError:
                break
            except Exception as e:
                print(f"Executor: Error during message sending: {e}")
```

danielSanchezQ: Yeah, something like that would do. Or in an `async for` using `asyncio.as_completed` if we need the result.

holisticode (commented):
Looks good to me, just a little nit

da/executor/dispersal.proto
```diff
@@ -0,0 +1,56 @@
...
+
+message DispersalRes {
+    oneof message_type {
+        bytes blob_id = 1;
+        DispersalErr err = 2;
+    }
+}
+
+message SampleErr {
+    oneof message_type {
```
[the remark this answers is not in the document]
bacv: I've added `oneof` for easier expandability later on. At the time I didn't think of another error that might happen during the sampling on the DA Node side of things, but if we'd come up with something in the future, no breaking changes would be introduced in the implementation.

bacv:
Merging this into the `da-poc` branch, subsequent adjustments will be added as new PRs.

PR #102, opened 2024-07-09
Cl: nullifier proof statement

davidrusu:
this adds a nullifier proof that is used to prove knowledge of the nullifier sk


cl/src/merkle.rs
```diff
@@ -67,15 +67,14 @@ pub fn verify_path(leaf: [u8; 32], path: &[PathNode], root: [u8; 32]) -> bool {
 }
 
 pub fn path<const N: usize>(leaves: [[u8; 32]; N], idx: usize) -> Vec<PathNode> {
```
zeegomo, line 69: nit: what's the advantage in hardcoding the merkle tree length?

goas/methods/nullifier/src/main.rs
```diff
@@ -0,0 +1,57 @@
+/// Nullifier Proof
+///
+/// Our goal: prove the nullifier nf was derived from a note that had previously been committed to.
+///
+/// More formally, nullifier statement says:
+/// for public input `nf` (nullifier) and `root_cm` (root of merkle tree over commitment set).
+/// the prover has knowledge of `output = (note, nf_pk, nonce)`, `nf` and `path` s.t. that the following constraints hold
```
zeegomo, line 7: don't you have to prove knowledge of nullifier sk for *input* notes? why is this called output?

cl/src/merkle.rs
davidrusu: mostly "because we can". We can drop later it if we find it is inconvenient for some reason.

goas/methods/nullifier/src/main.rs
davidrusu: Inputs are outputs of some other tx. Note commitments on chain are commitments to transaction outputs (recall UTXO stands for unspent transaction outputs.)

So here the prover is demonstrating that this input is derived from a valid output of some past transaction

PR #103, opened 2024-07-12
cl/integrate risc0

davidrusu:
- cl/ is now a workspace holding `cl`, `proof_statements`, `risc0_proofs` and `ledger`
- `cl::merkle::verify_path` is now `cl::merkle::path_root`
- `ledger` now is building the nullifier proof 
- `cl` tests are passing again


cl/risc0_proofs/nullifier/src/main.rs
```diff
@@ -0,0 +1,28 @@
...
+/// 2. note_cm = output_commitment(output)
+/// 3. verify_merkle_path(note_cm, root, path)
+use cl::merkle;
+use cl::nullifier::Nullifier;
+use proof_statements::nullifier::{NullifierPrivate, NullifierPublic};
+use risc0_zkvm::guest::env;
+
+fn main() {
+    let secret: NullifierPrivate = env::read();
+    assert_eq!(secret.output.nf_pk, secret.nf_sk.commit());
```
zeegomo, line 19: why take `nf_pk` as input if we recalculate it anyway to check consistency?

zeegomo (approved):

cl/cl/src/bundle.rs
```diff
@@ -24,11 +18,11 @@ pub struct BundleWitness {
     pub balance_blinding: Scalar,
 }
 
-#[derive(Debug)]
-pub struct BundleProof {
-    pub partials: Vec<PartialTxProof>,
-    pub balance_blinding: Scalar,
-}
+// #[derive(Debug)]
```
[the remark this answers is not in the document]
davidrusu: Ah, yep meant to remove these. All these proof structs and logic are moving into the ledger crate. This is to avoid a circular dependency where the risc0 proof depends on cl which in turn depends on risc0 for the proofs.

cl/ledger/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn it_works() {
+        let result = add(2, 2);
+        assert_eq!(result, 4);
+    }
+}
```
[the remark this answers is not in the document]
davidrusu: Yes, thanks for pointing it out.

cl/risc0_proofs/nullifier/src/main.rs
davidrusu: You're right, we don't need this assert

davidrusu: (The public output is the nullifier which is different from nf_pk)

PR #104, opened 2024-07-13
Mixnet: Use `max_message_size` and `max_path_length` for Sphinx encoding

youngjoon-lee:
We've defined two parameters: `max_message_size` and `max_mix_path_length`, but we haven't used it for Sphinx encoding because the `pysphinx` library didn't support those parameters.
Now that I modified the `pysphinx` to accept those parameters (instead of using hard-coded parameters), we can pass those parameters to `pysphinx`.

As a reminder,
- The size of Sphinx packet is determined by those two parameters. Those are global parameters shared to all nodes in the network.
- All Sphinx packets have the same size, even if an actual message in several packets is shorter than `max_message_size`.
- All Sphinx packets have the same size, even if an actual mix path of several packets is shorter than `max_mix_path_length`.
- If an user tries to use values longer than those parameters, `pysphinx` raises an exception.

### DIFF
--- a/mixnet/node.py
+++ b/mixnet/node.py
@@ -53,7 +53,9 @@ class Node:
         """
         A handler to process messages received via gossip channel
         """
-        sphinx_packet = SphinxPacket.from_bytes(msg)
+        sphinx_packet = SphinxPacket.from_bytes(
+            msg, self.global_config.max_mix_path_length
+        )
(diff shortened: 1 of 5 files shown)


youngjoon-lee:
Closing this with the same reason mentioned in https://github.com/logos-co/nomos-specs/pull/98#issuecomment-2238569126

PR #105, opened 2024-07-15
Mixnet: Simulation

youngjoon-lee:
This is the first working version of the NomMix simulation. 

Details can be found in the [documentation](https://www.notion.so/NomMix-Sim-Getting-Started-ee0e2191f4e7437e93976aff2627d7ce?pvs=4). 

When reviewing this PR, I would like to recommend you to focus on how the simulation works and what outputs it provides. Of course, I tried my best to implement details as clean as possible, but they are subject to change according to our design and next experiment goals. Thus, I don't recommend you to take too much time on reviewing all implementation details at this stage. I think it's easier to merge this first working version and extend it over and over, before making it too big.

- You will be able to understand the overall behavior of the simulation from the `config.ci.yaml` in this PR and the "documentation".
- Also, you can see what outputs the simulation provides from the "documentation". The shorter version is `README.md` in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,10 +19,12 @@ jobs:
         uses: actions/setup-python@v5
         with:
           # Semantic version range syntax or exact version of a Python version
-          python-version: '3.x'
+          python-version: "3.x"
       - name: Install dependencies
         run: pip install -r requirements.txt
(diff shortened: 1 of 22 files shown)


youngjoon-lee:
I’m moving this to the new repo: nomos-simulations

PR #106, opened 2024-07-16
CL: expand scope of nullifier proof to an "input proof"

davidrusu:
Input proofs ensure validity of:
- nullifiers
- balance commitments
- death commitments


zeegomo (approved):
Good work!

PR #109, opened 2024-07-19
Mixnet: Deprecate old spec

youngjoon-lee:
As discussed with Alvaro, I moved the existing old mixnet spec to the `deprecated/` directory because we are going to design the new mixnet protocol.

I will rebase the PR #98 once we merge this PR, so that it adds new mixnet spec files in the `mixnet/` directory.


deprecated/mixnet_v1/bls.py
```diff
@@ -2,7 +2,7 @@
 
 import blspy
 
-from mixnet.utils import random_bytes
+from deprecated.mixnet_v1.utils import random_bytes
```
alvatar, line 5: Is there a way to eliminate the deprecate.* prefix?
The Python environment should be living inside the deprecated/mixnet_v1, unaware of being deprecated. What do you think?

alvatar: We need to move carnot here as well, so I was hoping for a method that didn't require changing the code much

youngjoon-lee: > Is there a way to eliminate the deprecate.* prefix? The Python environment should be living inside the deprecated/mixnet_v1, unaware of being deprecated. What do you think?

I agree. We can use relative imports. I just pushed a commit for that: 1b4a860.

But, because we are using absolute imports for all code in this repo, if we want to move something to the `deprecated/` folder, we need to change the imports to the relative imports. If not, CI will be failed when running unit tests (also, Linter will complain). Of course, we can exclude the unit test of deprecated codes in the CI, but I think it's not a big deal to change the import paths.

youngjoon-lee: Plus, I don't think it's better to use relative imports for all other codes in this repo right now. As far as I know, using absolute imports is a recommended way, considering the case in which we publish this library to PyPI. 
So, I'd suggest to keep using absolute imports for live codes, and change import paths to relative if we want to deprecate something.

youngjoon-lee: If we don't want to run unit tests for deprecated code in the CI, and if we can ignore linter errors (such as `module XXX is not found`), I can revert the import path changes, and just move the directory to the `deprecated/`. That looks also good.

alvatar: Thanks Youngjoon. I think the arguments are solid.
However, I lean on _not maintaining CI tests for deprecated code_. What now works, may stop working at some point and then we will spend time just fixing those. Time that is completely wasted. I think deprecated code should be considered dead and is just left for reference.

You can leave the imports in any way that you see fit, but I think we should remove this from CI.

youngjoon-lee: I agree. I'm also leaning to that way: not running CI for deprecated code. So, I reverted all import path changes. Now, you can only see that files were moved without any change in this PR. They're excluded from the CI because we don't have the `deprecated/__init__.py`.

alvatar: Thanks. Do we need to do anything to change CI? @bacv

alvatar (approved):

deprecated/mixnet_v1/bls.py
youngjoon-lee: > Thanks. Do we need to do anything to change CI? @bacv

I think we don't. I checked that the test codes under the `deprecated/` folder are not run in CI.

alvatar: great, let's merge then

PR #113, opened 2024-08-28
Include PoL spec into cryptarchia specs

thomaslavaur:
Three tests fail right now because we unbonded the proof and the parent block. This can be done through cryptographic signatures and conflict resolution.


cryptarchia/cryptarchia.py
```diff
@@ -651,13 +662,13 @@ def phi(f: float, alpha: float) -> float:
...
-    return 1 - (1 - f) ** alpha
+    return alpha * (- log(1-f) - 0.5 * alpha * log(1-f) ** 2)
 
 
-class MOCK_LEADER_VRF:
+class LEADER_VRF:
     """NOT SECURE: A mock VRF function"""
 
-    ORDER = 2**256
+    ORDER = 2**253
```
davidrusu, line 671: where does this 253 come from again?

thomaslavaur: Yeah true I mixed the Circom implementation and the risc0 one. I'll remove this. (In Circom it's needed for comparison purpose if we operate over a prime field of 255 bits.)

cryptarchia/cryptarchia.py
```diff
@@ -445,20 +453,23 @@ def verify_slot_leader(
...
+        proof: LeaderProof,
         # coins are old enough if their commitment is in the stake distribution snapshot
         epoch_state: EpochState,
         # nullifiers (and commitments) are checked against the current state.
         # For now, we assume proof parent state and current state are identical.
         # This will change once we start putting merkle roots in headers
         current_state: LedgerState,
     ) -> bool:
+        threshold_0 = int(- (LEADER_VRF.ORDER * log(1 - self.config.active_slot_coeff)) / (epoch_state.total_active_stake()))
+        threshold_1 = int(- (LEADER_VRF.ORDER * log(1 - self.config.active_slot_coeff) ** 2) / (epoch_state.total_active_stake() ** 2))
```
davidrusu, line 465: this is used multiple times, let's factor these out into a function:

```
def lottery_thresholds(...):
    ...
    return (threshold_0, threshold_1)
```

cryptarchia/cryptarchia.py
```diff
@@ -173,28 +173,38 @@ def nullifier(self) -> Id:
...
 
-    def verify(self, slot: Slot, parent: Id):
-        # TODO: verification not implemented
-        return slot == self.slot and parent == self.parent
+    def verify(self, slot: Slot,
+               _epoch_nonce: Id,
+               _threshold_0: int,
+               _threshold_1: int,
+               _eligible_notes: set[Id],):
+        return slot == self.slot #TODO: and verification of the proof with risc0
```
davidrusu, line 207: We can write out the verification constraints in python. I.e. this serves as a spec for the NIZK statements.

PR #119, opened 2025-03-05
feat(cryptarchia): add cryptarchia v1 chain synchronization

youngjoon-lee:
# DRAFT

This PR will be opened after the [spec document](https://www.notion.so/Cryptarchia-v1-Chain-Synchronization-Protocol-1a48f96fb65c80468338dd1a8b26a14d) is updated accordingly.


cryptarchia/cryptarchia.py
```diff
@@ -592,6 +614,17 @@ def _infer_total_active_stake(
         )
         return int(prev_epoch.inferred_total_active_stake - h * blocks_per_slot_err)
 
+    def blocks_by_slot(self, from_slot: Slot) -> Generator[BlockHeader, None, None]:
+        # Returns blocks in the given range of slots in order of slot
+        # NOTE: In real implementation, this should be done by optimized data structures.
```
youngjoon-lee, line 619: I just realized that this can be done by iterating the `local_chain` and all `forks` backward. I'll update this function soon.

davidrusu (approved):
nice work, tests are thorough

PR #121, opened 2025-03-20
Cryptarchia/drop orphan proofs

davidrusu:


cryptarchia/cryptarchia.py
```diff
@@ -143,112 +161,68 @@ def encode_sk(self) -> bytes:
...
-        h = blake2b(digest_size=32)
-        h.update(b"coin-evolve")
-        h.update(self.encode_sk())
-        h.update(self.nonce)
-        evolved_nonce = h.digest()
-
-        return Coin(nonce=evolved_nonce, sk=self.sk, value=self.value)
-
-    def commitment(self) -> Id:
-        # TODO: mocked until CL is understood
```
youngjoon-lee, line 156: Is it safe to remove this TODO?

cryptarchia/cryptarchia.py
```diff
@@ -268,63 +242,42 @@ class LedgerState:
...
+    # set of note commitments
+    commitments: set[Hash] = field(default_factory=set)
 
-    # set of commitments eligible to lead
-    commitments_lead: set[Id] = field(default_factory=set)
-
-    # set of nullified coins
-    nullifiers: set[Id] = field(default_factory=set)
+    # set of nullified notes
+    nullifiers: set[Hash] = field(default_factory=set)
```
youngjoon-lee: it seems that we don't add anything to these lists. should it be done in the `apply` function below?

cryptarchia/cryptarchia.py
davidrusu: Yes, this is the correct commitment implementation now

cryptarchia/test_common.py
```diff
@@ -40,57 +39,53 @@ def on_block(self, block: BlockHeader):
...
+        content_id=Hash(b"CONTENT_ID", content),
+        leader_proof=MockLeaderProof(note, Slot(slot), parent=parent.id()),
     )
 
 
 def mk_chain(
-    parent: BlockHeader, coin: Coin, slots: list[int]
-) -> tuple[list[BlockHeader], Coin]:
+    parent: BlockHeader, note: Note, slots: list[int]
+) -> tuple[list[BlockHeader], Note]:
```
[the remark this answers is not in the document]
davidrusu: yes, good catch

cryptarchia/cryptarchia.py
youngjoon-lee: Sorry, this is not straightforward for me to understand. Could you please give me an example of a transaction that produces new commitments and nullifiers?

davidrusu: these would be transactions in the block contents. we abstract over that in this implementation since we only deal with block headers.

With UTXO's, every transaction produces nullifiers of the input notes that were consumed and generates commitments for the output notes.

All nullifiers and commitments across all transactions in a block will be added to the commitment / nullifier sets here. (our blocks are empty here so we thats why you are not seeing this)

The reason why we are not updating these sets anymore in PoL is because with PoLv2, we no longer reveal the nullifier of the winning note. (in PoLv1, we did and that forced us to evolve the note to avoid linking past PoL wins together) PoLv2 now proves that it has not spend it's note by proving that the note is not a member of the nullifier set.

youngjoon-lee: Thanks for detailed explanations. Very helpful.
So, the nullifiers, which are produced by transactions, can be added to the `LedgerState.nullifiers` without privacy concern because those nullifiers are not the result of generating a leader proof. 
Anyone can be a leader if one of its notes wins a slot lottery and if it is not in the `LedgerState.nullifiers`. Instead of revealing the nullifier of the winning coin, the leader provides a proof of non-membership.
I think now I understand this correctly :) ?

davidrusu: yes, that's all correct!

PR #126, opened 2025-05-20
Update DA encoder/verifier to v1.1

megonen:


thomaslavaur (commented):
Looks good to me. Can be merged when SHA is removed :p

da/test_encoder.py
```diff
@@ -33,24 +34,26 @@ def assert_encoding(self, encoder_params: DAEncoderParams, data: bytes):
...
-            for i, (chunk, proof) in enumerate(zip(row, proofs)):
-                self.assertTrue(
-                    kzg.verify_element_proof(bytes_to_bls_field(chunk), commitment, proof, i, ROOTS_OF_UNITY)
-                )
-
-        # verify column aggregation
-        for i, (column, proof) in enumerate(zip(encoded_data.extended_matrix.columns, encoded_data.aggregated_column_proofs)):
-            data = DAEncoder.hash_commitment_blake2b31(commitment)
+        h = DAVerifier._derive_challenge(encoded_data.row_commitments)
+        com_C = encoded_data.row_commitments[0]
```
danielSanchezQ, line 41: Lets be explicit
```suggestion
        `column_commitment` = encoded_data.row_commitments[0]
```

da/test_encoder.py
```diff
@@ -33,24 +34,26 @@ def assert_encoding(self, encoder_params: DAEncoderParams, data: bytes):
...
+        power = h
+        for com in encoded_data.row_commitments[1:]:
+            com_C = com_C + com * int(power)
+            power = power * h
+
+        for i, (column, proof) in enumerate(zip(encoded_data.extended_matrix.columns, encoded_data.combined_column_proofs)):
+            v = BLSFieldElement(0)
```
danielSanchezQ, line 48: what is `v`?

da/test_encoder.py
```diff
@@ -33,24 +34,26 @@ def assert_encoding(self, encoder_params: DAEncoderParams, data: bytes):
...
+            power = BLSFieldElement(1)
+            for chunk in column.chunks:
+                x = BLSFieldElement(int.from_bytes(bytes(chunk), byteorder="big"))
```
danielSanchezQ, line 51: What is `x`?

da/verifier.py
```diff
@@ -1,106 +1,70 @@
...
+          h = BLAKE2b-31( DST || bytes(com1) || bytes(com2) || ... )
+        """
+        h = blake2b(digest_size=31)
+        h.update(_DST)
+        for com in row_commitments:
+            h.update(bytes(com))
+        digest31 = h.digest()  # 31 bytes
+        # pad to 32 bytes for field element conversion
+        padded = digest31 + b'\x00'
+        return BLSFieldElement.from_bytes(padded)
```
danielSanchezQ, line 43: We have this defined in encoder right? Better to move it to `common`

danielSanchezQ (changes requested):
Looking good. I would just go another round of fixing naming of things. I know we come from academia where math symbols are used (`x`, `y`, `h`, etc) but it is usually better readable if we actually say what is each thing. Its also much better for devs when they have to implement the thingy 😉

megonen:
> Looking good. I would just go another round of fixing naming of things. I know we come from academia where math symbols are used (`x`, `y`, `h`, etc) but it is usually better readable if we actually say what is each thing. Its also much better for devs when they have to implement the thingy 😉

You’re right, you know this is the first try! I’ll get used to implementation step by step 😄  I’ll make the changes you suggested and clarify the naming 👍

da/test_encoder.py
megonen: "combined_commitment". I take the first row commitment, then multiply the others with the powers of h, and add them together. I mean like this com_C=com_0+h*com_1+h^2*com_2+... I will update the naming

da/test_encoder.py
megonen: combined evaluation point. Actually, each chunk element is taken to compute the combined_evaluation_point for the corresponding column. I had denoted this as v in the cryptographic protocol. x represents the BLS element form of each chunk. I’ll update these parts accordingly

da/kzg_rs/poly.py
```diff
@@ -49,11 +49,19 @@ def __sub__(self, other):
...
+                self.modulus
+            )
+        elif isinstance(other, Polynomial):
+            result = [0] * (len(self.coefficients) + len(other.coefficients) - 1)
+            for i in range(len(self.coefficients)):
+                for j in range(len(other.coefficients)):
+                    result[i + j] = (result[i + j] + self.coefficients[i] * other.coefficients[j]) % self.modulus
+            return Polynomial(result, self.modulus)
+        else:
+            raise TypeError(f"Unsupported type for multiplication: {type(other)}")
```
danielSanchezQ, line 64: Better to have a custom method for one of the branches. Otherwise it is difficult to read. Or at least, separate each branch into methods.

da/test_encoder.py
```diff
@@ -33,24 +35,26 @@ def assert_encoding(self, encoder_params: DAEncoderParams, data: bytes):
...
-
-        # verify column aggregation
-        for i, (column, proof) in enumerate(zip(encoded_data.extended_matrix.columns, encoded_data.aggregated_column_proofs)):
-            data = DAEncoder.hash_commitment_blake2b31(commitment)
+        h = derive_challenge(encoded_data.row_commitments)
+        combined_commitment = encoded_data.row_commitments[0]
+        power = h
+        for com in encoded_data.row_commitments[1:]:
+            combined_commitment = bls.add(combined_commitment,bls.multiply(com, power))
+            power = power * h
```
danielSanchezQ, line 46: Letters are free! 💵 
```suggestion
        for commitment in encoded_data.row_commitments[1:]:
            combined_commitment = bls.add(combined_commitment,bls.multiply(commitment, power))
            power = power * h
```

da/verifier.py
```diff
@@ -1,106 +1,55 @@
...
-        # Ensure idempotency: Implementers should define how to avoid redundant verification.
-        return True
+        # 1. Derive challenge
+        h = derive_challenge(blob.row_commitments)
+        # 2. Reconstruct combined commitment: combined_commitment = sum_{i=0..l-1} h^i * row_commitments[i]
+        combined_commitment = blob.row_commitments[0]
+        power = h
+        for com in blob.row_commitments[1:]:
+            combined_commitment = bls.add(combined_commitment,bls.multiply(com, power))
+            power = power * h
```
danielSanchezQ, line 45: Easier to read, explicit better than implicit.
```suggestion
        for commitment in blob.row_commitments[1:]:
            combined_commitment = bls.add(combined_commitment,bls.multiply(commitment, power))
            power = power * h
```

da/encoder.py
```diff
@@ -66,71 +65,36 @@ def __rs_encode_row(row: Row) -> Row:
...
+            int_pow = (int_pow * h_int) % BLS_MODULUS
+        return combined
+
+    def _compute_combined_column_proofs(self, combined_poly: Polynomial) -> List[Proof]:
+        total_cols = self.params.column_count * 2
         return [
-            kzg.generate_element_proof(i, polynomial, GLOBAL_PARAMETERS, ROOTS_OF_UNITY)
-            for i in range(len(column_commitments))
+            kzg.generate_element_proof(j, combined_poly, GLOBAL_PARAMETERS, ROOTS_OF_UNITY)
+            for j in range(total_cols)
```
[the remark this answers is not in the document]
megonen: In the cryptographic spec, the i’s were representing rows and the j’s columns, that’s why I used j 😄 but I can change

da/kzg_rs/poly.py
megonen: I had made this change because I was getting an error when multiplying the polynomial by a constant value. But now I found a way to fix it without touching this part, so I’ll remove this change anyway

danielSanchezQ:
After battling with the intermediary mess I got to have everything clean. @megonen & @thomaslavaur take a last look and we can have this green 😄

danielSanchezQ (approved):

PR #127, opened 2025-05-28
Add Boostrapping/Online modes

zeegomo:
Add Boostrapping and Online modes to cryptarchia, including relevant tests. The Boostrap mode uses the Genesis fc rule, while Online uses Praos. Swtitching between the two rules is left to the implementation and is specified in the public Notion as linked in the comment

### DIFF
--- a/cryptarchia/cryptarchia.py
+++ b/cryptarchia/cryptarchia.py
@@ -1,5 +1,5 @@
 import functools
-import itertools
+from itertools import islice
 import logging
 from collections import defaultdict
 from copy import deepcopy
@@ -322,6 +322,7 @@ class Follower:
         self.ledger_state = {genesis_state.block.id(): genesis_state.copy()}
         self.epoch_state = {}
         self.state = State.BOOTSTRAPPING
+        self.lib = genesis_state.block.id()  # Last immutable block, initially the genesis block
 
     def to_online(self):
         """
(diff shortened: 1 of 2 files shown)


cryptarchia/cryptarchia.py
```diff
@@ -317,12 +321,29 @@ def __init__(self, genesis_state: LedgerState, config: Config):
...
 
     def validate_header(self, block: BlockHeader):
         # TODO: verify blocks are not in the 'future'
         if block.parent not in self.ledger_state:
             raise ParentNotFound
 
+        if not is_ancestor(self.lib, block.parent, self.ledger_state):
+            # If the block is not an ancestor of the last immutable block, we cannot process it.
+            raise ImmutableFork
+
```
davidrusu, line 346: Isn't the `block.parent not in self.ledger_state`  check above sufficient?

We prune the block tree so that all blocks that we hold are descendents of LIB, if the parent is in our block tree, then the parent is a descendent of LIB, therefore this block is a descendent of LIB.

davidrusu (approved):
Mainly the question about whether we should also purge bocks along side forks, I think we should, no point in keeping that state around, and it allows us to remove the ancestor check in the block validation logic

cryptarchia/cryptarchia.py
davidrusu: oh right, good point.

So maybe the check should be
1. parent is in the block tree
2. parent height >= LIB height -- parent is forking at or after LIB

davidrusu (approved):

